PR #5, opened 2018-01-05
1 hour = 3600 seconds, not 3700

bls1999:
Also updated epic everything comment from rank tokens to ee


jacob-grahn:
Haha, I had the extra seconds in there to give people a little bit of a bonus. I’ll accept the comment change, but I’d like to keep the time at 3700. Thanks!

bls1999:
Right on! I figured but just wanted to make sure, since you're famous for nevr making typos. :)

PR #13, opened 2018-01-06
One *last* tiny change for pr2_moderation.php

bls1999:
Re-enabled privilege to kick any user from Fred, just in case. Affected line: 12.

So now it is: The player kicking has to be a temp mod, and also has to be either kicking someone less than a moderator or be Fred (4291976).

This is the last one for this file!  I promise!

### DIFF
--- a/server/pr2_moderation.php
+++ b/server/pr2_moderation.php
@@ -9,7 +9,7 @@ function kick($socket, $data){
 	$player = $socket->get_player();
 	
 	// if the player actually has the power to do what they're trying to do, then do it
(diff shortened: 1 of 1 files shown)


bls1999:
Line 12.

So now it is: The player kicking has to be a temp mod, and also has to be either kicking someone less than a moderator or be the server owner (as defined in $guild_owner).

jacob-grahn:
That's pretty cool, I like it. It's just needs "global $guild_owner;" at the top of the function to get access to that global variable. I'll go ahead and add that part in.

PR #19, opened 2018-01-08
Fixes

bls1999:
- Server doesn't crash on demod anymore, line 215.
 - Add an offline/account DNE error, lines 191-194.


jacob-grahn:
Pushing this live, let's see if it works!

PR #20, opened 2018-01-08
Last 2 small fixes for pr2_moderation.php

bls1999:
Affected lines: 211-219

### DIFF
--- a/server/pr2_moderation.php
+++ b/server/pr2_moderation.php
@@ -209,13 +209,13 @@ function demote_moderator($socket, $name) {
 
 	if($from_player->group == 3){
 		$to_player = name_to_player($name);
+		if(isset($to_player) && $to_player->group < 2) {
+			$from_player->write('message`Error: '.$name.' is not a moderator.');
+		}
 		if(isset($to_player) && $to_player->group == 2) {
 			$to_player->group = 1;
 			$to_player->write('setGroup`1');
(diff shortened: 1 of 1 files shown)


server/pr2_moderation.php
```diff
@@ -209,14 +209,14 @@ function demote_moderator($socket, $name) {
 
 	if($from_player->group == 3){
 		$to_player = name_to_player($name);
+		if(isset($to_player) && $to_player->group < 2) {
+			$from_player->write('message`Error: '.$name.' is not a moderator.');
+		}
 		if(isset($to_player) && $to_player->group == 2) {
```
jacob-grahn: Should this be "else if"?

bls1999: It could be. I can change that

jacob-grahn: Oh, I see. Both statements can't be true at the same time. I'm just trying to figure out how this stops two messages from getting sent?

bls1999: If it's left as it is, this is the execution:

 - Demod
 - Check for a player's group and echo an error message if they're not a mod

Which produces two messages:
 - ____ has been demoted.
 - Error: ___ is not a moderator.

So, by changing the order we will prevent the double check and just go with one message.

jacob-grahn: Gotcha, thanks!

PR #22, opened 2018-01-08
mod pro/demotion commands fixes and security

bls1999:
@jacob-grahn 

First of all, fixed the command paths.  As of now, the code on the master repo isn't searching in the /commands folder.  I am unable to promote or demote on Isabel at all.

Now, for the bulk of the changes.

The theory that's operating here is that I can pass the promoter as an entire player object from pr2_moderation.php to promote_to_moderator.php and demod.php through the shell arguments.  Then, the error messages use the caught exceptions already built into the command phps.  The caught exception function is expanded a bit in both demod and promote_to_moderator.php.  If nothing is caught, then the script continues to write a success message to the promoter (dubbed $admin) by doing ``$admin->write("message`$name has been promoted to a $type moderator!");``

The only problem I can think of is that the shell exec() wouldn't be able to figure out what the heck $admin->write means.  In that case, some globals would probably need to be added to both of the command phps.  We could also go the route of eliminating the shell commands altogether and making promote_to_moderator.php and demod.php functions instead.

At any rate, I doubt it's anything that will be too difficult to crack.  See db84d82 for some more details.

### DIFF
--- a/server/commands/demod.php
+++ b/server/commands/demod.php
@@ -3,54 +3,73 @@
 
 require_once(__DIR__ . '/../fns/db_fns.php');
 
-$port = $argv[1];
-$user_name = $argv[2];
-$admin = $argv[3]; // player passed as argument for use with confirmation/error messages
-$caught_exception = false;
+function demote_mod($port, $user_name, $admin, $demoted_player) {
 
-// if the user isn't an admin, kill the script
-if($admin->group != 3) {
-	echo $admin->name." lacks the power to demote. Quitting...";
-	$admin->write("message`Error: You lack the power to demote $user_name.");
-	exit;
-}
-
-try {
-	$connection = user_connect();
-	$user_id = name_to_id($connection, $user_name);
-	$safe_user_id = addslashes($user_id);
-
-
-	//delete mod entry
-	$result = $connection->query("DELETE FROM mod_power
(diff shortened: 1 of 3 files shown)


bls1999:
Also, does throw new Exception stop code execution in promote_to_moderator.php?

jacob-grahn:
My bad with the shell file paths. 

The reason for the shell commands is to offload database queries from the main loop. This server works in a constant loop. Read, write, read, write, read, write, etc. Anything that stopped the loop, like waiting for a query to finish, would cause lag for everyone playing.

A *much* better solution exists today. It's possible to run async queries that won't stop the game loop. Check out MYSQLI_ASYNC on this page: http://php.net/manual/en/mysqli.query.php

bls1999:
That sounds good.  Let me push some more commits that takes care of this.

Would you be opposed to completely eliminating the commands folder and just integrating the functions into pr2_moderation.php?

bls1999:
Right on.  Gonna work on this some more.  Don't merge it just yet.

bls1999:
Okay, should be good now.  Check to make sure that:
 - The DB calls are structured as they should be in both demod and promote_to_moderator.php
 - Everything is called correctly
 - You have mysqlnd installed
 - Echo exception (at the end of demod and promote_to_moderator.php) isn't just a relic of the past

PR #31, opened 2018-01-14
Removed extra htmlspecialchars validation

bls1999:
The validation would've caught every message and flagged it as containing an illegal character, since data is passed to send_chat as ``chat`playername`playergroup`message``.

I *think* I figured out how to pull without adding 10 million commits in one.  Here's hoping.


bls1999:
HOORAY!  Only one commit!

jacob-grahn:
Looks good to me!

PR #32, opened 2018-01-16
fix server crash in promote_to_moderator.php

bls1999:
`$result` should've been `$admin_result` on line 34.
`$result` should've been `$user_result` on line 50.
Both have been changed so the server doesn't crash.

Please remove the limits on moderator promotions (1 every 6 hours) so that we can test the new promotion/demotion mechanisms.

### DIFF
--- a/server/fns/promote_to_moderator.php
+++ b/server/fns/promote_to_moderator.php
@@ -31,7 +31,7 @@ function promote_mod($port, $name, $type, $admin, $promoted_player) {
 									WHERE user_id = '$safe_admin_id'
 									LIMIT 0,1",
 									MYSQLI_ASYNC);
-	$admin_row = $result->fetch_object();
+	$admin_row = $admin_result->fetch_object();
 
 	//check for proper permission in the db (3rd + final line of defense before promotion)
 	if($admin_row->power != 3) {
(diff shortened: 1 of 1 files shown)


jacob-grahn:
Looking at this script, the MYSQLI_ASYNC flag is going to cause us some problems. None of the lines that check the result of a query will actually work, because the script won't wait for the query to finish before moving on to the next line. For now, I'm going to take the easy path and remove the MYSQLI_ASYNC flags. That'll cause some lag whenever someone is promoted to a moderator, but oh well, that won't happen very often.

You mentioned that there is a rate limit on moderator promotions, which probably means a script on pr2hub is involved. It may be possible to move this logic to pr2hub, so it doesn't slow down the game server. 

I'll upload the files from pr2hub to this repo soon, I just need to go through it all to make sure there won't be any passwords that get uploaded...

jacob-grahn:
The test code is running on Isabel again, and I changed the mod rate limit to 60 seconds.

bls1999:
@jacob-grahn

The rate limit on mod promotions is actually in the server code under promote_to_moderator.php.  The server checks for recent promotions in the database and then proceeds with the promotion if it's all good.  I don't think there's any pr2hub scripts involved, although I could be wrong.

Oh boy, pr2hub files!  Please don't release any passwords. :P

PR #39, opened 2018-01-16
Hat awarding issue

EternalUF:
When users logged in during New Years they received the Santa hat, this will give them the appropriate Party hat!


jacob-grahn:
Looks good to me, thanks Eternal and bls!

PR #43, opened 2018-01-17
Add html prevention in exception message

Myst1cZ:
An small fix on exception to only allow new line tags.
Edit: added another new line only allowed on show_record.php

### DIFF
--- a/http_server/www/bans/show_record.php
+++ b/http_server/www/bans/show_record.php
@@ -97,7 +97,7 @@ if($is_mod === true) {
 
 
 echo "<p>$html_mod_name banned $html_banned_name for $f_duration on $formatted_time.</p>
-		<p>Reason: $reason</p>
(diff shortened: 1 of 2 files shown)


bls1999:
Again, this is not needed. The only person that would be able to see that message is the user.

Myst1cZ:
Added only new line tags allowed in ban reason at show_record.php

jacob-grahn:
Thanks Myst1cZ!

jacob-grahn:
It looks like in recent versions of php just using `<br>` will work for `<br>` and `<br/>`

http://php.net/manual/en/function.strip-tags.php

> Note:
> In PHP 5.3.4 and later, self-closing XHTML tags are ignored and only non-self-closing tags should be used in allowable_tags. For example, to allow both `<br>` and `<br/>`, you should use:
> ```php
> <?php
> strip_tags($input, '<br>');
> ?>
> ```

PR #45, opened 2018-01-17
Update for index

Antonhs30000:
Added Artifact hint, fixed links and added _blank, also better css, image for arti: https://i.imgur.com/IsFJdnD.png


bls1999:
I like this, but I wonder if we can add this to artifact_hint.php instead of putting it on the homepage.

Antonhs30000:
Putting it on the homepage makes it faster. It's also possible to add a reload button or automatically reload the hint after 1 minute using JavaScript.

bls1999:
The point, though, is that not everyone would enjoy having it on the front page. Plus, it's very easy to do /hint in the chat anyway.

jacob-grahn:
Cool idea! I'd be open to having a page at hint.php, and linking to it from the bottom of index.php.

Antonhs30000:
Oh yeah, you're right! Haha @bls1999

PR #46, opened 2018-01-17
CSS for the index.php

Antonhs30000:
This is the CSS for the new index.php


jacob-grahn:
Hey Antonhs,

I set this up at pr2hub.com/hint.php, and it's using hint.css. When I was taking stuff out, I think I messed up the styling though!

PR #47, opened 2018-01-18
Fixed XSS exploit

Myst1cZ:
@jacob-grahn 
Fix for XSS exploit (could be used to steal login tokens in some cases, that would allow logging in as targeted player without password by entering the token cookie to own browser.)


jacob-grahn:
Looks good, thanks!

PR #48, opened 2018-01-18
Fixed Guild Exploit

Antonhs30000:
This code can prevent users from using illegal characters to disable other user accounts.

### DIFF
--- a/http_server/www/guild_create.php
+++ b/http_server/www/guild_create.php
@@ -8,7 +8,6 @@ try {
 	$note = find( 'note' );
 	$guild_name = find( 'name' );
 	$emblem = find( 'emblem' );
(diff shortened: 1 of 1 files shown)


jacob-grahn:
What if we use `preg_match` instead? I think it would work like this with your regex:

```php
if (preg_match('/[^a-zA-Z0-9#-.:;=?@~! ]/', $guild_name) === 1) {
  throw new Exception('There is an invalid character in your guild name. '
  .'The allowed characters are a-z, 1-9, and !#$%&()*+.:;=?@~.');
}
```

This way we don't have to deal with a `$noexploit` variable

Antonhs30000:
Sounds cool! c:

PR #50, opened 2018-01-18
Rework mod nav mechanic, index.html -> php, css updates

bls1999:
- Moved mod nav determination to output_fns.php
 - Removed mod nav in individual files
 - Converted index.php to index.html
 - Tweak Antonhs' css changes


jacob-grahn:
The only thing I'm concerned about is replacing the index.html which can be served very quickly with a php file that does some db work. 

I think `check_moderator` throws an error if the user is not a moderator. There is an `is_moderator` function in fns/data_fns.php that does what you're looking for. It does it in a silly, silly way, but whatever. I was young at the time.

bls1999:
@jacob-grahn

jacob-grahn:
Just a heads up, I changed this around a little bit. 

For this line in output_fns.php to work, it needs a database connection:
```
$db = new DB(); \\ added this line...
$formatting_for_mods = check_moderator($db, false);
```

But then I realized that would create two database connections per request, so I got rid of that and had the $formatting_for_mods variable get passed in. We loose the automatic awesomeness that your pull request had, but it works without opening two connections.

PR #51, opened 2018-01-18
Create terms-of-use.php

bls1999:
Make sure everything in here is to your liking.


jacob-grahn:
Thanks for keeping it short but sweet. :+1:

PR #52, opened 2018-01-18
Fix guild exploit, typos

bls1999:


jacob-grahn:
Looks good, thanks bls!

bls1999:
Good catch on the dot! My bad.

PR #53, opened 2018-01-18
Replace index.php require_once with .htaccess

bls1999:
In the directories:
 - bans
 - mod

Also, added .htaccess to backups, since it was missing one.

### DIFF
--- a/http_server/www/backups/.htaccess
+++ b/http_server/www/backups/.htaccess
@@ -1,0 +1,1 @@
+DirectoryIndex backups.php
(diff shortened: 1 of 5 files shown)


bls1999:
Since this won't work with nginx, here's a guide on how to specify directory indexes:
https://www.keycdn.com/support/nginx-directory-index/

I can try to do this, however I have 0 experience with nginx so I really don't want to screw anything up.

bls1999:
@jacob-grahn found another resource that may come in handy:

https://www.digitalocean.com/community/questions/htaccess-config-to-nginx
https://isenselabs.com/posts/how-to-install-and-run-opencart-on-nginx-server

jacob-grahn:
I'm not too concerned about the index.php files being there. I figure the less server config I have to do if pr2 ever needs to switch servers again, the better.

PR #55, opened 2018-01-18
Fix friends/ignored server check from 2012 update

bls1999:
Lines 61-63.

On user lists (friends, ignored), it used to say "Derron" or "Carina" before the status language was updated in 2012.  Now, it says "Playing on Derron", which cuts off the server name on most names.  This change will fix that.


jacob-grahn:
Fixing bugs from 2012... what history

PR #57, opened 2018-01-18
Fixed artifact message

bls1999:
This should make it so it only displays after a user has gotten the artifact first. Check my work, please.


Antonhs30000:
By the way, the other changes were mine. (Replaced \n to <br> and safe user name). :p @jacob-grahn

jacob-grahn:
Thanks guys, hopefully this fixes it

PR #58, opened 2018-01-18
Add action logging to more mod actions

bls1999:
Added action logging support for:
 - Deleting guilds
 - Lifting bans
 - Banning
 - Unpublishing levels
 - Deleting guilds
 - Archiving PM reports

Check my code and make sure it all works with `$db->call`, whether it's on the http server or the actual PR2 servers.

### DIFF
--- a/http_server/www/ban_user.php
+++ b/http_server/www/ban_user.php
@@ -1,4 +1,5 @@
 <?php
+
 require_once('../fns/all_fns.php');
 
 $banned_name = find('banned_name', '');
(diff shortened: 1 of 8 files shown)


bls1999:
There are quite a few commits in this one, but most of it is just me correcting mistakes I made.  No need to be worried. :) @jacob-grahn

jacob-grahn:
It looks like `htmlspecialchars` is used when the logs are being displayed in mod/action_log.php. I don't think we need to call `htmlspecialchars` before inserting the data into the db.

bls1999:
@jacob-grahn Oops. I meant to call `mysqli_real_escape_string`, but it looks like that isn't needed in `$db->call` either.  I'll make another pull.  This does require edits from you to work, though.

PR #60, opened 2018-01-18
Fix broken loader when accessing pr2hub over https

bls1999:
This pull is an extension of #50 

Through a .htaccess redirect, the mixed content error is defeated and the pr2 loader and client are both able to be successfully loaded over https and http.


jacob-grahn:
Interesting! pr2hub is running on nginx, so htaccess files won't work. A similar nginx config should accomplish the same goal though. I'll look into it.

PR #62, opened 2018-01-18
Fixed HTML in guild_top.php [FAIL]

Antonhs30000:


Myst1cZ:
I thought doing this earlier aswell but i wasn't just sure since htmlspecialchars converts quotes aswell to &quot; and that response gives JSON which has lots of quotes, wouldn't this affect them too?

bls1999:
Won't work.  You'd be encoding the entire reply, so the JSON characters would get replaced too.  I don't think putting htmlspecialchars inside json_encode will work either, since $reply is not a string.  As far as I know, htmlspecialchars only works with strings.

Myst1cZ:
@bls1999 Would addslashes have any effect? It should add slash before next one but not sure if it would prevent html though.

bls1999:
@Myst1cZ nope, addslashes is only useful for php and MySQL requests. It won't escape html characters.

Myst1cZ:
@bls1999 Was kind of expecting that, it was just thought i had since i've seen pr2 pms have slashes on the pms if checking through get_messages.php where it shows 2 slashes instead of one and in pr2 there's only one slash in message.

bls1999:
@antonhs30000 If you were looking for the option, you can close the pull my clicking the close button next to the comment button.

PR #63, opened 2018-01-18
Just an html prevention

Myst1cZ:
Just adding an html prevention to forgot password
Edit: added fix for HTML in guilds_top.php (will still give valid JSON but in text format instead of html)

### DIFF
--- a/http_server/www/guilds_top.php
+++ b/http_server/www/guilds_top.php
@@ -1,5 +1,9 @@
 <?php
 
+if (!headers_sent()){
+	header("Content-Type: text/plain");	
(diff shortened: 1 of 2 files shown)


jacob-grahn:
Awesome, thanks Myst1cZ

http_server/www/guilds_top.php
```diff
@@ -1,5 +1,9 @@
 <?php
 
+if (!headers_sent()){
```
bls1999: Correct me if I'm wrong, but doesn't every request send headers?

PR #66, opened 2018-01-19
Key changes to first artifact finder mechanic

bls1999:
This pull contains some key changes to the first artifact finder mechanic.  The script will now:

 - Echo a prize box that says "Bubble Set" and has the bubble body as an icon
 - There's too much going on when the user finds the artifact first.  Instead of popping up a congratulatory message, PM it to the user (check my code on line 45 to make sure I'm using `$db->call` correctly)
 - Award the prize via gain_part() (taken from Player.php) and auto-set the prize upon return to the lobby

Of course, check my code, but this is a lot simpler than the code was before since it uses functions that already exist.  Plus, it actually works. :)

LMK if/when you merge this code and get it running on Isabel so I can test if it works for awarding it to the first finder.

### DIFF
--- a/server/fns/artifact_first_check.php
+++ b/server/fns/artifact_first_check.php
@@ -9,14 +9,38 @@ function artifact_first_check($port, $player) {
 	$safe_user_name = htmlspecialchars($player->name);
 
 	try {
+		
 		$first_finder = $db->grab( 'first_finder', 'artifact_find', array($user_id) );
 
 		if( $first_finder == $user_id ) {
-			award_parts( $db, $user_id, 'head', [27] );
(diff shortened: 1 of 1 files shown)


Myst1cZ:
@bls1999 does this show prize popup when entering level with artifact on it or just when finishing with artifact?

bls1999:
@Myst1cZ just when finishing with the artifact. Sort of like the Jigg Hat win mechanic; a prize box at the end but not at the start of the race.

Myst1cZ:
@bls1999 Ok, that's awesome.

jacob-grahn:
This is awesome, indeed. I'll get this running on Isabel now.

PR #67, opened 2018-01-19
Update download.txt

Grovbi:
Updated to the current build. Kong doesn't use this link, but I know BubbleBox at least does.


bls1999:
Good catch!

BubbleBox actually forces https on their site now, which inadvertently breaks the PR2 loader.  I have no idea why; it just does.

jacob-grahn:
Thanks Grovbi! I forgot that file even existed...

PR #69, opened 2018-01-19
check for only spaces in guild names

bls1999:
For both files, I used `strlen(trim($guild_name)) === 0` on line 47.


jacob-grahn:
Ok, cool

PR #71, opened 2018-01-19
Some more changes and fixes

Myst1cZ:
1. Added missing image, width and height to image, exception message and changes to CSS file for hint page.
2. Added text/plain type for get_messages.php
3. Added "-- artifact hint --" text image to artifact hint page.


Myst1cZ:
Note: img/artifact.png image was missing so i added it.

Myst1cZ:
@bls1999 It's kinda just recommended to use the headers_sent part to make sure that the header isn't sent after client has already received headers, it just means sending extra stuff with no use if directly using the header function.

Antonhs30000:
Artifact image has to be transparent.

Antonhs30000:
Also, this works as well. 
artifact

bls1999:
https://cdn.discordapp.com/attachments/249678944956055562/378292633501696010/260898610734956574.png

I have the right padding set to 40px to make it look more centered on the page.  I was thinking that the page could say -- Artifact Hint --, and then under it there could be the artifact picture, and then under that would be the actual hint data.

I'll also integrate this with the functions in output_fns.php later so there's less repeated code.

Myst1cZ:
@bls1999 Thanks for the transparent artifact image, i'll get into this now.

jacob-grahn:
Thanks, this page is looking really good! [https://pr2hub.com/hint.php](https://pr2hub.com/hint.php)

PR #74, opened 2018-01-19
Fix welcome/winner PMs, which display the literal text \r instead of a new line

bls1999:
Changed \r to real new lines.

Couple of questions @jacob-grahn:
 - I assume the default color for each part is set in the client?  If not, could we set the bubble body to black?  It sort of blends in with the prize popup.
 - Will this award the bubble set even if the first finder has it already?  Or does gain_part check for the user having the part first?

Other than that, everything is working perfectly!

### DIFF
--- a/http_server/www/login.php
+++ b/http_server/www/login.php
@@ -145,7 +145,15 @@ try {
 		$db->call('pr2_insert', array($user_id));
 
 		//--- send them a welcome pm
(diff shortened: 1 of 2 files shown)


Antonhs30000:
The default color of each part is set in the client. 
image

bls1999:
Well, that sucks.  I might just want to set the icon for the prize pop-up as the artifact.  That might be a better move.

PR #75, opened 2018-01-19
Revert: Pull #63

bls1999:
Reverts jacob-grahn/platform-racing-2-server#63

Reason being: html is not parsed in guilds_top.php, but rather in the client.  htmlspecialchars would work better, but I don't know where to put it.

### DIFF
--- a/http_server/www/forgot_password.php
+++ b/http_server/www/forgot_password.php
@@ -14,7 +14,7 @@ $safe_email = addslashes($email);
 try{
 
 	if(!valid_email($email)){
-		throw new Exception("'" . htmlspecialchars($email) . "' is not a valid email address.");
+		throw new Exception("'$email' is not a valid email address.");
 	}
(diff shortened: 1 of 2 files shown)


Myst1cZ:
@bls1999 Umm.. ever viewed the guilds_top.php in browser? It renders html on the page due to it being in html format and not plain text, you are reverting exploit fix here.

Myst1cZ:
Extra note: this never was supposed to fix html inside pr2 but to fix html being parsed in the page on browser.

bls1999:
my bad

woodenstix:
I don't know if I can take you seriously xDD

Revert fix for exploit? New low for bls

bls1999:
@myst1cz Sorry, I read the pull wrong.  Didn't realize it wasn't intended for the client.

Myst1cZ:
@bls1999 Np, btw, i sent invite on discord to talk about fixes.

PR #81, opened 2018-01-20
Added protection against html parsing in client

bls1999:
This fix was relatively simple compared to what I was expecting. I was able to simply add a line that converts `$guild->guild_name` to `htmlspecialchars($guild->guild_name)`.  I also made sure the file sends characters in UTF-8 to prevent encoding errors.

Affected lines: 4, 35

### DIFF
--- a/http_server/www/get_player_info_2.php
+++ b/http_server/www/get_player_info_2.php
@@ -1,6 +1,6 @@
 <?php
 
-header("Content-type: text/plain");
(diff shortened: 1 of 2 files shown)


bls1999:
@Myst1cZ that htmlspecialchars fix not only fixes the raw response (makes it `&lt;` for instance), but it also parses in-game, making it readable as the intended characters.

jacob-grahn:
Looks good to me!

PR #86, opened 2018-01-21
Add jiggmin2.com to trusted urls

bls1999:
I solemnly swear that I am up to no good.

(It's a Harry Potter reference. I promise I'm only "up to" good things. Pinky promise.)


jacob-grahn:
jiggmin2.com has a forum on it where bad links like `pr2hub.com/ban_everyone.php` could be posted. I think it would be better if we listed specific urls like `https://jiggmin2.com/games/platform-racing-2`

bls1999:
Sounds good.  I've also found a way to edit the client, so I might want to add that link to the list somewhere down the road.  But that's a while away.

PR #88, opened 2018-01-21
Create .htaccess

Antonhs30000:
I made it so if you want to go to a page, you don't need to include .php 

Example: http://pr2hub.com/login.php ->http://pr2hub.com/login


jacob-grahn:
That's pretty cool, but the server is running nginx rather than apache. htaccess files will just sit there and eat donuts while doing nothing!

PR #91, opened 2018-01-21
Replaced some text in login.php

Antonhs30000:
Ex: Incorrect Username -> Login detailts are not correct.


isokissa3:
Why are you treating username like its secret? Everyone knows your username, pointless

PR #92, opened 2018-01-22
Fix ban editing and lifting

bls1999:
Fixed ban editing and lifting by correcting the file paths in /bans/show_record.php.
Affected lines: 90, 92, 95, 114-115


jacob-grahn:
These links miiiight need to have a `../` in front of them, but I'll see if it works

jacob-grahn:
Sweet, it works like magic!

PR #94, opened 2018-01-23
Misc security/housekeeping updates

bls1999:
Gerry (@Myst1cZ) decided to make a script that automatically deletes people's PMs because he's angry he's not getting unbanned for DDoSing and abusing exploits.  This pull should fix that exploit in most places.

 - Added updates to security when using GET as a request method (5218712, d5e3609, 6a8e0bc, 66c720e, 6b249f7, abf305b, 11ad5b2, 858cad4)
 - For security, tell the user both emails in the confirmation email (9228290, 5529330)
 - Deleted useless or duplicate phps (02ad2f1, 068a7ff, 7338f5b, a5c2c84)
 - Add swear filter for guild info (b04ee04, 8b98178, 6a8e0bc)
 - Added ban dispute link to banned popup (1388561)
 - Fixed delete all PMs bug where "uploading..." would never go away (4939b09)


jacob-grahn:
`use_login_token($db, $token);` doesn't check if the account is banned, which might be a problem. It might be better to have an option in token_login to require the token to be present in the GET variables. I'll disable the "delete all messages" option in the short term.

bls1999:
Gotcha.  I can fix this shortly.  Let me make a new pull.

bls1999:
@jacob-grahn actually, could we just add a banned check to `use_login_token($db, $token)`?

bls1999:
Wow okay I added functions that were already there in a few files. Smart, bls. Real smart. I'm just going to make a new pull.

PR #95, opened 2018-01-23
Security/filter updates

bls1999:
Fix exploits, filter swears.  Revision of pull #94


Myst1cZ:
@bls1999 Late reply on what you said in previous pull request: this wasn't because being angry but just for having fun.

Myst1cZ:
Note: there's an fix missing for one of the exploits i used.

Myst1cZ:
Just released source code of the exploits that went around jiggmin2:
https://github.com/Myst1cZ/ZeroPM

bls1999:
@myst1cz well, considering deleting people's PMs was the only thing that actually worked, this fix is sufficient.

Issue #99, opened 2018-01-23
F@H Prizes not awarding


bls1999:
Could this be because the code in `http_server/cron/award_folding_prizes.php` is enclosed in

/* these tags */

?

If so, there's a fix in #103.

PR #103, opened 2018-01-23
[potentially] fix folding awards

bls1999:
Potential fix for #99.  Just uncommented the entirety of http_server/cron/award_folding_prizes.php.


jacob-grahn:
Honestly I commented it out because I was thinking about removing it altogether. But since someone noticed it is non longer working, I guess there are people still using it.

jacob-grahn:
There are a few other things I'll have to do to get this working, now that I think about it!

PR #114, opened 2018-01-24
PR2Hub Player Search System

Antonhs30000:
Image: https://i.imgur.com/AOnVhk9.png

There's /view command, but it would be cool to have a PR2Hub Player Search System :wink:


bls1999:
Yup, this is the same situation as the guild script; this is my code.

PR #115, opened 2018-01-24
PR2Hub Guild Search System

Antonhs30000:
Image: https://i.imgur.com/K6jTxmJ.png

There's /guild but again, it would be cool to have a PR2Hub guild search system! ^^


bls1999:
You know, I was thinking this looked an awful lot like a script I coded.

And then it dawned on me.  This is the script I coded.

PR #117, opened 2018-01-25
Fixes for mod navs, action log, admin tools

bls1999:
Note on update_guild.php: the script says "error=A guild already exists with that name." when trying to update.  Since that was the specified error text, I can't really trust that that is the real error. I would assume the issue stems from the `$db->call` function not being able to handle the extra arguments.

Have a look and see what you can find.  d1a4950

### DIFF
--- a/multiplayer_server/fns/promote_to_moderator.php
+++ b/multiplayer_server/fns/promote_to_moderator.php
@@ -1,9 +1,13 @@
 <?php
+
 require_once(__DIR__ . '/db_fns.php');
+
 function promote_mod($port, $name, $type, $admin, $promoted_player) {
 	global $db;
+	
 	// boolean var for use in if statement @end
 	$caught_exception = false;
+	
 	// if the user isn't an admin on the server, kill the function (2nd line of defense)
 	if($admin->group != 3) {
 		$caught_exception = true;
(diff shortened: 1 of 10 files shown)


multiplayer_server/fns/promote_to_moderator.php
```diff
@@ -114,13 +131,24 @@ function promote_mod($port, $name, $type, $admin, $promoted_player) {
...
+			foreach (range(0,$server_count) as $server_id) {
+				$server_port = $servers->servers[$server_id]->port;
+				if ($port == $server_port) {
+					$server_name = $servers->servers[$server_id]->server_name;
+					break;
+				}
+			}
+			
 			// log action in action log
-			$db->call('mod_action_insert', array($admin->user_id, "$admin_name promoted $promoted_name to a $type moderator from $ip on port $port", $admin->user_id, $ip));
```
jacob-grahn: `$db->call` escapes all of the values passed to it. Doing it manually here will double escape the values. It doesn't matter much here, but if an admin had the name bo'b, they would come out of the db with the name bo\'b

jacob-grahn:
Alright, I added `owner_id` as an input to `guild_update`. I took out the `member_count` input, and set up the `guild_update` procedure to auto-calculate that value.

bls1999:
Perfect! It also seems that the emblem is being erased whenever I try to edit a guild.

PR #118, opened 2018-01-25
Fix tournament mode

bls1999:
This should fix tournament mode and prevent it from crashing the server and/or disconnecting the player.

Remember ya functions, Jiggy!


jacob-grahn:
Thanks!

Issue #121, opened 2018-01-25
Platform Racing 2 +0 exp 


Jmarroqu:
@jacob-grahn Since you're allowing macroing would it be okay if you can get rid of the +0 exp after simming/macroing twice in a sim map? This would help for us not to spam sim/macro maps and not to deal with the Pr2 Staff "Level spamming". Take a look here; http://prntscr.com/i5g9ac Thanks Jiggy :)

FearsPR2:
I agree with Colind. There'd be no need for publishing 9 pages of 6 sim levels each anymore.

woodenstix:
hahaha since macroing is legal now then simming should have no restrictions

MagicalBoss:
Yes! PLEASE FIX THIS ISSUE IT IS VERRY ANNOYING and makes macroing frustrating if I mess it up I gotta restart the process

PR #122, opened 2018-01-25
Fix for issue #121

fourst4r:
farewell sim detector


jacob-grahn:
Hmm, let me think on this one for a bit. If we do take this out, I'd like to replace it with something that helps the average-joe-not-simmer get a reasonable amount of exp, too.

woodenstix:
@jacob-grahn increased exp drops in campaign maps depending with how many people you're racing with

Jmarroqu:
The members of Gods of Glory were talking about a way for you can make it fair for racer and we had an idea @jacob-grahn I hope you like it; Maybe all Campaign maps should have a bonus exp for every defeated player they beat for example If you win 1st in a campaign map you'll get the exp you'll get + 150 exp then 2nd place would get +100, 3rd would get 50 and last place would get 5 or 10 exp and in return the +0 exp is gone.

VipaPR2:
That's actually a pretty good idea @Jmarroqu

LP-PR2:
I like @Jmarroqu 's idea, but I thought of something else that could benefit the campaign racers, and that's to increase the amount of exp you earn for winning duplicated parts, as nowadays 5 exp will not benefit anyone for anything, as you need 4 players to earn the parts, i think 50-100 exp for winning the part seems a lot more reasonable.

woodenstix:
remember playing PR3 in 2012? Sim maps didn't work, because you would get 0 exp from it (you didn't move).

jacob-grahn:
Thanks fourst4r, I went ahead and merged this in. We'll work on giving racers more exp next!

PR #124, opened 2018-01-25
Fix message archiving, housekeeping

bls1999:
This pull is a bunch of small fixes that come together to form a normal-sized fix.

Fixes to PM archiving: there was a stray `}` that was breaking the script. I removed that and it should work now. (51f1644)

Duplicate Files:
 - report_message.php = message_report.php (5c04bed)
 - delete_message.php = message_delete.php (2b2e076)
 - get_messages.php = messages_get.php (6544298)
 - send_message.php = message_send.php (a3af7d9)

I deleted the first of each file similarity above. The ones deleted are the ones not being used by the PR2 client.

Finally, 6af9d70 adds a check for an already reported message (since the custom error is currently broken).


jacob-grahn:
Looks great, thanks for cleaning those files up!

PR #126, opened 2018-01-26
Definitively fix emblem deletion, more housekeeping

bls1999:
7422130 should make it so editing a guild will preserve the emblem.  The scope of the function got me the first time, then it was not defining a variable earlier in the function but calling it at that time. 
 17th time's the charm, right?

Also in this pull:
 - Hopefully fix ban_edit.php's checkboxes (54d4c71)
 - Fix action logging in ban_edit.php (039d1ec)
 - http->https in ban_edit.php (e065091)
 - Fix broken redirect link in do_lift_ban.php (f96bae8)
 - New language in README.md (9b5a627)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,15 @@
 # Platform Racing 2 Server
 
+## Guidelines for Contributors
+
+Welcome to the PR2 GitHub repo!
+
+All contributions are welcome; However, please keep these things in mind when contributing:
+- Only edit things if you know exactly what they do and what you're doing.
+- If you want to edit something but don't exactly know what to do, post an issue and someone will assist you.
+- When submitting a pull request, leave a summary of your changes in the title/description.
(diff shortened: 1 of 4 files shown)


jacob-grahn:
Just a thought, what we use a function like 

```php
function checked ($value) {
  if ($value) {
    return 'checked="checked"'
  }
  else {
    return ''
  }
}
```

Then using it would look like 

```
$ip_checked = checked($ban->ip_ban);
```

bls1999:
That might work, although the value isn't passed as true/false, it's passed as 1/0.  If we were to go about that, it might be better to do something like this?

```function checked ($value) {
  if ($value === 1) {
    return 'checked="checked"';
  }
  else {
    return '';
  }
}```

(don't forget your semicolons!)

PR #127, opened 2018-01-26
Fix message reporting

bls1999:
I broke it :(
This fixes it :)

### DIFF
--- a/http_server/www/message_report.php
+++ b/http_server/www/message_report.php
@@ -20,11 +20,12 @@ try{
 	
 	
 	//make sure the message isn't already reported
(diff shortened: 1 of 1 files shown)


jacob-grahn:
:)

I think you were on the right track though. Pulling all of the data from the rows is slightly inefficient, just pulling the count seems better. Here's some incomplete and probably broken code!

```
$result = $db->query("SELECT COUNT(*) as count FROM whatever");
$row = $result->fetch_object();
if ($row->count > 0) {
  // already reported
}
```

bls1999:
I can go ahead and make this a new pull!  I tried to leave out "as count" and that screwed everything up I think.

PR #128, opened 2018-01-26
Fixes to action logging in promote_to_moderator.php and demod.php

bls1999:


woodenstix:
hope you dont break mod functions now

bls1999:
@woodenstix that's @jacob-grahn's job! :P

It don't think it will break anything, although I've been wrong before!

bls1999:
Will be combined with #133

PR #141, opened 2018-01-28
Server crash fixes when using commands in levels, new commands/edits

bls1999:
- Fixes server crashes when using /give, /a, /promote, /clear and any chat effects in levels.
 - Doesn't give an error when using functions that a player doesn't have permission to use.
 - Players/server owners can now check tournament status by using /t status, using `tournament_status()` in fns/announce_tournament.php
 - Add emote :shrug:
 - Add /promote for admins (same thing as /give, except the text is "bls1999 has promoted " instead of "bls1999 has given ")
 - Update /help command

### DIFF
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -245,7 +245,12 @@ class Player {
 	public function send_chat($chat_message) {
 		global $guild_owner;
 		global $player_array;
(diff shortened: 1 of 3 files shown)


jacob-grahn:
¯\_(ツ)_/¯

PR #145, opened 2018-01-29
Add backtick check to login.php, fix country code

bls1999:
Adds a sanity check to all logins that makes sure backticks aren't used in usernames.
Also uses an IP lookup API to determine the country code for player_deep_info.php.


jacob-grahn:
Well ok, let's give keycdn.com a try!

PR #147, opened 2018-01-30
Actually fix server crashes

bls1999:
I usually get it on the second try.

 - Cleaned up and reworked the dynamics for issue_tournament.php
 - Actually possibly maybe fixed server crashes (see commit 6111c4a for more info)

### DIFF
--- a/http_server/www/mod/action_log.php
+++ b/http_server/www/mod/action_log.php
@@ -13,7 +13,7 @@ try {
 	$db = new DB();
 	
 	//make sure you're a moderator
-	$mod = check_moderator($db);
+	$mod = check_moderator($db, false);
 
 	//get actions for this page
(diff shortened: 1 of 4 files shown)


jacob-grahn:
Here's the error that crashed Isabel when I was looking at it:

```
Player->send_chat('/give a')
Undefined property: Player::$chat_room Player.php on line 336
Fatal error: Call to send_chat on null Player.php on line 336
```

bls1999:
Just thought of something better.  What if we define a variable called `$player_chatroom` using two if statements and then call the variable later down the line?  That way, the check is already done.

I can add this later!

jacob-grahn:
Cool, that should work. I dunno if there will ever be a situation where both chat_room and game_room exist, or neither of them exist, but it would probably be good to check for those edge cases too.

PR #148, opened 2018-01-30
A very little fix at player_search.php

Antonhs30000:
Added one more `<br />`, also @jacob-grahn please upload the image (-- Player Seach --) to PR2Hub

https://i.imgur.com/AOnVhk9.png


bls1999:
Here's an idea: what if we embed the ttf file of gwibble and then write all the text in that font using `<font face="Gwibble">your text here</font>`?  No need for a bunch of images, and we get as much gwibble text as we want.

jacob-grahn:
Ha, we can add gwibble font to the roadmap

PR #149, opened 2018-01-30
Update Game.php

Grovbi:
I think this is a leftover from the cat captcha. Also includes a possible fix for an issue(?) involving random parts not showing up in single player matches that appeared around early 2014-ish. Regardless if it was intentional or not, it'd be worth restoring given how few players there are nowadays.


jacob-grahn:
It does look like that function isn't used any more, thanks for cleaning it out.

PR #150, opened 2018-01-30
Optimize code, fix /t error and /clear crash

bls1999:
Added game/chat room check and put in the variable `$player_room`.
Also cleaned up issue_tournament.php a bit.

The only issue I'm having is trying to figure out how to throw just the exception text I want to throw.  I have the `throw new Exception("text here!")` statements defined in /fns/issue_tournament.php. Then, I put the function in a `try{}` block and caught it in a `catch{}` block, both in Player.php.  When I was testing the stat error, I got this:

```
Error: Exception: Stat values can only be between 0 and 100. in /home/jiggmin/platform-racing-2-server/multiplayer_server/fns/issue_tournament.php:52
Stack trace:
#0 /home/jiggmin/platform-racing-2-server/multiplayer_server/Player.php(316): issue_tournament('/t none 1000 10...')
#1 /home/jiggmin/platform-racing-2-server/multiplayer_server/client/lobby.php(118): Player->send_chat('/t none 1000 10...')
#2 /home/jiggmin/platform-racing-2-server/multiplayer_server/socket_server.php(132): client_chat(Object(pr2_server_client), '/t none 1000 10...')
#3 /home/jiggmin/platform-racing-2-server/multiplayer_server/socket_server.php(160): pr2_server_client->handle_request('260
```

Further testing showed that all exceptions thrown produced the same result.

I'm about 73.86% sure that this pull won't fix that monstrosity, but it'll provide some other fixes and put me one step closer to figuring out what the heck is causing it.

EDIT: Fixed it in commit 55c9e52.  I was calling `$e` instead of `$e->getMessage()`

### DIFF
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -342,8 +342,9 @@ class Player {
 						}
 						catch (Exception $e) {
 							$caught_exception = true;
+							$err = $e->getMessage();
 							$err_supl = " Make sure you typed everything correctly! For help with tournaments, type /t help.";
-							$this->write('systemChat`Error: ' . $e . $err_supl);
+							$this->write('systemChat`Error: ' . $err . $err_supl);
 						}
 						
 						// if an error was not encountered, announce the tournament to the chatroom
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -246,6 +246,24 @@ class Player {
 		global $guild_owner;
 		global $player_array;
 		
+		// find what room the player is in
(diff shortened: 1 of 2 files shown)


jacob-grahn:
`$e` gives you a different error message than `$e->getMessage()`??

bls1999:
Apparently so! From what I could find, `$e` is sort of an array and is basically the equivalent of `var_dump($e)`. Therefore, `getMessage()` is only part of it.  I don't really get it either but I guess it'd be useful for some. :man_shrugging:

PR #151, opened 2018-01-30
Fixes to /admin directory, mod_log.php

bls1999:
mod_log.php was missing a semicolon that was killing the entire script, so I went ahead and fixed that.
Each commit has specific changes, but in summary: I fixed a bunch of stuff.

I'm going to edit set_campaign.php to accept arguments for custom prize lists as well.  In the end, though, I need you to enable the database to accept a request to set the campaign.  Or, now that I think about it, would the database function from the old Campaign of the Fortnight setter still work?  If so, could you send me the code for that?  Thanks!


jacob-grahn:
Thanks! I'll look around in the db and see if there's any procedure to set the campaign.

PR #136, opened 2018-01-28
Player Search Page for PR2Hub

Antonhs30000:
I just coded a Player Search page for PR2Hub! ^^

Tell me if there are possible exploits! :3

Image ( --Player Search-- ): https://i.imgur.com/AOnVhk9.png

PR #152, opened 2018-01-31
Optimization & Fixes

bls1999:
This is mostly related to #136 and #148.

 - Move the css files to their own directory (www/style)
 - Instruct output_fns.php to detect what the page name is.  If it finds the artifact hint page, call hint.css.  Else, call pr2hub.css.
 - Add gwibble font to www/style, make a css file to format it (gwibble.css), and call it in all files to enable image-free gwibble on all pages.
 - Update hint.php to use the new output functions


jacob-grahn:
Now I want to write some really long and important document in gwibble

PR #156, opened 2018-02-01
Fixes to PM reporting, admin tools, new functions in data_fns.php

bls1999:
- Hopefully finally fix PM reporting
 - Fix action logging in a bunch of files
 - Some cosmetic edits and housekeeping
 - Added two new functions (`is_empty()` and `check_value()`; you can see what they do in 7e4ddd1)


jacob-grahn:
`// if it's a month/year ban coming from PR2, correct the weird ban times`
LOL

PR #157, opened 2018-02-01
Add guild de artifact to the vault

bls1999:
Same as Fred and Ghost parts, except it's with the artifact.

This was in the code for PR2 as placeholder text, but I noticed you never added it to the vault.  Seems like it might attract some players and earn you some money in the process!

This pull is totally optional.  I don't mind at all if you close it because you don't want people to be able to buy access to the artifact hat, although I do want to mention that it's a prize for an ongoing contest.  The ones who win don't abuse it (and can't really; the time is 30 seconds) and really enjoy the challenge of trying to complete an entire level in 30 seconds.


jacob-grahn:
Cool idea! Thanks!

PR #158, opened 2018-02-01
Allow admins to restart a PR2 server

bls1999:
This is another completely optional pull.

I've been a long-time supporter of staff being able to restart servers since they've been known to be wonky every now and then.  If you don't want to give admins that power, that's totally fine and I completely understand; However, if you aren't vehemently opposed to the idea, I believe it would be a good resource for admins.

The process is simple: admin types `/restart_server` and gets prompted with a big scary message that looks like this:

```
WARNING: You just typed the server restart command. If you choose to proceed, this action will disconnect EVERY player on this server. Are you sure you want to disconnect ALL players and restart the server? If so, type: /restart_server yes, i am sure!
```
Then, if the admin types `/restart_server yes, i am sure!`, the action is logged in the admin log and the server restarts.

### DIFF
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -245,6 +245,7 @@ class Player {
 	public function send_chat($chat_message) {
 		global $guild_owner;
 		global $player_array;
+		global $port;
 
 		// find what room the player is in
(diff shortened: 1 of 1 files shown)


jacob-grahn:
Sounds good to me. I'll get this rolling on Isabel

PR #167, opened 2018-02-03
Update Prizes.php + Others

Grovbi:
Optional thing because I'm weird: reverted a few names of stuff on prize pop-ups to their original pre-2014 ones. Most guides still call them these, so why not? *shrugs*

Edit: Added another commit to make the epic Jump Start Hat universal like the other campaign hats. (Right now it functions like a normal campaign prize; needs 4 players and only winner gets it.)

Edit 2: Added a few more.

### DIFF
--- a/http_server/www/admin/part_ids.php
+++ b/http_server/www/admin/part_ids.php
@@ -70,7 +70,7 @@ try {
 36 - Reindeer
 37 - Crocodile
 38 - Valentine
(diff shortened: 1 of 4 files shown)


bls1999:
I also have a pull on the way that cleans up the Prizes.php file a bit!  This will be added to my edits.

jacob-grahn:
Ok, cool!

PR #173, opened 2018-02-05
Create staff.php | finally the correct one :p

Antonhs30000:
This is the correct one, merge that, it has function inside and I fixed a typo :p @jacob-grahn

### DIFF
--- a/http_server/www/hint.php
+++ b/http_server/www/hint.php
@@ -4,8 +4,7 @@ require_once('../fns/output_fns.php');
 
 output_header('Artifact Hint');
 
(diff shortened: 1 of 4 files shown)


bls1999:
I have some fixes I'd like to submit for this. Stay tuned.

Myst1cZ:
I see aswell few things that would be good idea to change

Antonhs30000:
Alright, Ben! ^^

bls1999:
okay @jacob-grahn, this is good to merge now.

PR #182, opened 2018-02-09
Really fix the epic Jump Start Hat

Grovbi:
I think the game thought "true" was a description, oops. That or it just hates custom campaigns.


bls1999:
yeah this will fix it.

jacob-grahn:
Thanks Grovbi!

PR #183, opened 2018-02-10
Add time left command in Player.php

Stxtics:
This is a command for owners of Private Servers. When they type /timeleft or /ps it will tell them when their private server will expire (I think). This is my first ever time doing something in php but I believe this should not cause any errors. The only issue I think that could occur is how this expire time might be shown. From what I've seen I think it is in a date format but I am not 100% sure. Anyone can modify this if they want but if you can tell me why you are changing what you did that will help. Thanks!

### DIFF
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -438,6 +438,7 @@ class Player {
 					$this->write('systemChat`This command cannot be used in levels.');
 				}
 			}
+			// restart server command
 			else if (($chat_message == '/restart_server' || strpos($chat_message, '/restart_server ') === 0) && $this->group >= 3) {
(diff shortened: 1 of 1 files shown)


bls1999:
@Stxtics since I am submitting a change to Player.php, let me add this to my pull so Jiggmin doesn't have to fix the merge conflicts.

Stxtics:
@bls1999 do I have to like allow you to do that or will it let you?

bls1999:
I'm just going to add it manually to my pull because that seems like the easier course of action.

bls1999:
Did it https://github.com/bls1999/platform-racing-2-server/blob/patch-1/multiplayer_server/Player.php

PR #187, opened 2018-02-11
Fix future possible folding issues in register_user.php

Stxtics:
Users that contain #, %, and @ cannot be folded. I believe it's due to how folding at home stores usernames with those characters and this causes PR2 not to find a match with the usernames to be able to award the tokens and hats. 

Because of this I changed the allowed characters in usernames although % wasn't there so I'm not sure how to make it not allowed. Also 0 is allowed in usernames and technically the invalid character message tells you it is not.

### DIFF
--- a/http_server/www/register_user.php
+++ b/http_server/www/register_user.php
@@ -31,10 +31,10 @@ try{
 	if(is_obsene($name)){
 		throw new Exception('Do try to think of a different name.');
 	}
-	$test_name = preg_replace("/[^a-zA-Z0-9#-.:;=?@~! ]/", "", $name);
+	$test_name = preg_replace("/[^a-zA-Z0-9-.:;=?~! ]/", "", $name);
 	if($test_name != $name){
 		throw new Exception('There is an invalid character in your name. '
-							.'The allowed characters are a-z, 1-9, and !#$%&()*+.:;=?@~.');
+							.'The allowed characters are a-z, 0-9, and !$&()*+.:;=?~.');
 	}
 
 


Stxtics:
Actually I think folding at home just doesn't allow users with those characters to be folded.

bls1999:
Yeah, we could try this; However, I think a better fix would be to update the F@H script to test `_` replacements.

Stxtics:
But #, % and @ are not allowed in folding @ home usernames? So are you saying for people who want to fold accounts that have that in their username should put _ instead and then the fah script will check for that instead?

bls1999:
Yeah, that's how it's been since the game started.  Try searching for b_by*G on the folding list.

Stxtics:
Ok I kinda see what you mean now but I still think it's a pain that if you make an account with #, % or @ in the name you will have to put _ instead when u fold it (if the fah script is changed).

bls1999:
Oh it's definitely a pain, although I *think* F@H automatically replaces those characters with `_`.

bls1999:
What script drives that bot?  Are you using the F@H api?
This is what PR2 uses to determine if you have enough F@H points: http://fah-web.stanford.edu/teamstats/team143016.html <http://fah-web.stanford.edu/teamstats/team143016.html>

Stxtics:
http://fah-web2.stanford.edu/cgi-bin/main.py?qtype=userpage&username={username}&teamnum=14301

{username} is username entered with command.

jacob-grahn:
The hard part of trying to match `some_name` with `some%name` is that there could also be an account called `some@name`. We could award the parts to every match to be on the safe side, but people would figure that out pretty quickly and make multiple matching accounts on purpose...

jacob-grahn:
@Stxtics It looks like the `%` was already being filtered, it just wasn't in the description. Thanks for the pull request!

yayeet1337:
Okay but when do I get prizes? I folded 3 tokens on 22th January and they still havent arrived :(

bls1999:
Assuming F@H updates daily, you should get them within 24 hours.

PR #188, opened 2018-02-13
staff.php fixes

bls1999:
Fix the broken columns
Also adds a link to the staff list on player_search.php if the user is a mod or more


Antonhs30000:
Really?

PR #190, opened 2018-02-13
Add/fix debugging, other misc fixes/changes

bls1999:
There seems to be an issue retrieving some variable or accessing the shutdown_server() function in the /restart_server command.  This adds some debugging code to make sure we can figure out exactly what is causing this error.

This should also fix the /timeleft command, and the confirmation/error messages for successful temp mod promotions and unsuccessful demotions.

### DIFF
--- a/http_server/www/account_change_email.php
+++ b/http_server/www/account_change_email.php
@@ -25,54 +25,63 @@ try{
 	$new_email = $data->email;
 	$pass = $data->pass;
 
-	//--- connect
+	// connect
 	$db = new DB();
 
-	//--- check thier login
+	// check their login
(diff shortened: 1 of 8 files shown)


jacob-grahn:
The error message from /restart_server is:

Notice: Undefined variable: db in Player.php on line 460
Fatal: Call to a member function call() on null in Plyaer.php on line 460

bls1999:
@jacob-grahn thanks for letting me know.  I would've never guessed that.  Hopefully my most recent commits will fix it.

jacob-grahn:
Trying to catch up on these PRs :+1:

PR #191, opened 2018-02-13
Disable HH on tournament, formatting in HappyHour.php

bls1999:
This SHOULD disable HH on servers with tournament mode enabled... but do check my code to make sure it works.


bls1999:
That should work!  I actually had it in the isActive function originally but was unsure about variable scope.

jacob-grahn:
Oops, we committed the change at the same time

bls1999:
¯\_(ツ)_/¯

PR #194, opened 2018-02-14
Add missing new line in player_search.php and fix staff.php

Stxtics:
If you view a staff member using player_search.php it currently looks like this. https://i.imgur.com/1CuUEjV.png


bls1999:
Fixed the safe_name bug in #190

Stxtics:
yeah whoops i checked all of them except that one

jacob-grahn:
Thanks, this looks good!

PR #196, opened 2018-02-15
Add support for guild owners to transfer ownership of their guilds to other users

bls1999:
A lot of people were asking for this.

These scripts will allow guild creators to transfer ownership of their guild to another user.  They would go to the url https://pr2hub.com/guild_transfer.php and enter their email, password, and the name of the new guild owner.  They would then get an email confirmation with a code (like email changes).  Finally, upon clicking the link, the owner id is changed to the new user.

It's important to note:
 - Form submissions to guild_transfer.php must go in POST, thereby disabling XSS/CSRF attacks (theoretically; I'm not a security expert by any means)
 - Even if someone was able to perform an XSS/CSRF attack, the password/email of the user would be missing, rendering it useless
 - All ownership changes are logged in the same way email changes are (requires some db setup from you).

Let me know if you have any questions or comments.  I think this will be widely appreciated.


jacob-grahn:
Hey Bls, good PR! This will take some database work and some testing, so I'll work on this when I have an evening free.

PR #201, opened 2018-02-16
Create email search tool for admins

bls1999:
Allows admins to search for users by email address.  Hopefully in the future I'll figure out a way to grab the username from different email changes too.  That way, not only will we be able to search by the current email address, but we'll also be able to search by any address that has ever been associated with an account.

### DIFF
--- a/http_server/www/admin/search_by_email.php
+++ b/http_server/www/admin/search_by_email.php
@@ -32,6 +32,7 @@ catch (Exception $e) {
 	output_header('Error');
 	echo "Error: $message";
 	output_footer();
+	die();
 }
 
 // admin validated try block
--- a/http_server/www/admin/search_by_email.php
+++ b/http_server/www/admin/search_by_email.php
@@ -1,0 +1,103 @@
+<?php
+
+require_once('../../fns/all_fns.php');
+require_once('../../fns/output_fns.php');
+
+$email = find_no_cookie('email', '');
+$group_colors = ['7e7f7f', '047b7b', '1c369f', '870a6f'];
+
+// this will echo the search box when called
+function output_search($email='', $incl_br=true) {
+	echo "<form name='input' action='' method='get'>";
+	echo "Email: <input type='text' name='email' value='$email'>&nbsp;";
+	echo "<input type='submit' value='Search'></form>";
+	if ($incl_br) {
+		echo "<br><br>";
+	}
+}
+
+// admin check try block
+try {
+	
+	//connect
+	$db = new DB();
+	
+	//make sure you're an admin
+	$admin = check_moderator($db, false, 3);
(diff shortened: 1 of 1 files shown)


woodenstix:
yeah yeah BLS! PLEASE GIVE US ABILITY TO FIND PPL BY EMAIL SO I CAN HIJACK THEIR EMAILS AND RECOVER ACCOUNTS YEES!

bls1999:
@woodenstix lol read the first sentence of the pull request

woodenstix:
Yes, yes, give admins more info so that when we get a corrupt admin they can leak everything! Thanks, bls1999!

bls1999:
1. Admins already have the ability to see users' email addresses, which would mean a theoretical corrupt admin could already leak that data.
2. Using your logic, we should not allow mods or admins to see IP addresses or prior bans since they can be leaked.
3. This could be used to help find lost accounts for people.

bls1999:
Also, in the interest of full disclosure, @woodenstix's original message was:

`yeah yeah BLS! PLEASE GIVE US ABILITY TO FIND PPL BY EMAIL SO I CAN HIJACK THEIR EMAILS AND RECOVER ACCOUNTS YEES!`

And my response was:

`Read the first sentence of the pull request`

woodenstix:
Your response didn't even matter, I edited it 15 seconds later. Are you trying that I am somehow wrong by stating a problem? Keep lying.

woodenstix:
Actually I deleted it :/

Antonhs30000:
Well done Ben! Keep the good work! c:

bls1999:
@jacob-grahn this one is also a quick merge, since it doesn't use any db functions

jacob-grahn:
I'm a little bit uneasy about email addresses being accessible in the first place. I never really wanted to add them in, but people needed a password recovery feature...

With the way companies collect everyone's personal info (and then loose it in a data breach), I'd like us to be as much on the other end of the spectrum as possible.

There are valid and good reasons for this pull request, but if it's possible to get by without access to emails I think that would be better.

Anywho. I'm open to discussion if you disagree.

EternalUF:
Hey Jiggmin! The issue currently, which sparked the creation of this pull is it is extremely difficult figuring out the number of accounts an email has associated to it, so when it comes to mass ban evaders or when someone just forgets the accounts they have on their email (given they provide us the email first). I believe in the future bls1999 wants to implement a search by IP, which would also greatly benefit us for similar reasonings, but I believe he stuck with this first since it was quicker to accomplish. The thing is, the things in this pull is already possible, but it would take an extremely long time and highly inefficient clogging the servers with manually requests member by member. This just eases everything. But once again, I’m open to all opinions!

PR #214, opened 2018-02-22
Fix update_guild.php after guild transfer merge

bls1999:
change `guild_transfer_select` to `guild_transfer_select_by_code`.

Also, it looks like I've already added the code for requester IP and confirmer IP to guild_transfer.php and confirm_guild_transfer.php.  Can you make this information available to admins by calling `guild_transfers_select_by_guild` in guild_deep_info.php?  This would make the information we see pretty much identical to the email change information.


jacob-grahn:
I added request_ip and confirm_ip to the guild_transfers_select_by_guild procedure. Thanks!

PR #215, opened 2018-02-22
Update delete_old_accounts.php

VipaPR2:
I thoroughly agree with the idea of a "script" which cleans out inactive/unused accounts however I'd like to suggest some small changes to the algorithm which is used to determine whether an account is unused. As of right now when executed the script collects a list of users from the database which have been inactive for a time length greater than three years, an account is then deleted if it meets the following requirements; has not yet reached rank ten and a total level play count which has not reached fifty. Personally I feel these values could be increased and still reflect the definition of an unused/inactive account, therefore my suggestion is to increase the level required to twenty as this would also coincide with the rank required to create a guild and I also feel inclined to suggest an increase of the minimum play count to five hundred or a thousand as fifty is a very measly amount and was more often than not easily achieved just by uploading a single level once. These changes aren't exactly a priority nor a necessity however I feel like myself and several if not many other community members would like to see this change as it would result in the release of more usernames which had been taken in the past and rarely used.

### DIFF
--- a/http_server/cron/delete_old_accounts.php
+++ b/http_server/cron/delete_old_accounts.php
@@ -26,7 +26,7 @@ while($row = $result->fetch_object()) {
 	}
 
 	$str = "$i $user_id plays: $play_count rank: $rank.";
-	if($play_count > 50 || $rank > 10) {
+	if($play_count > 1000 || $rank > 20) {
 		output("$str SPARE");
 	}
 	else {


woodenstix:
https://grahn.io/feedback/
https://grahn.io/feedback/

bls1999:
Hey @VipaPR2, can you post this on JV2's PR2 Suggestions forum?  More details in README.md on this repo.

(I'd be happy to take a look and see what I can improve)

Grovbi:
This would delete a lot of levels, so I'm against this. Remember, levels would only stay on Newest for very short times back when the game was popular. Not even most mod levels from around the time have more than 300 plays. Combine this with the fact that most users from around then were below rank 20 since EXP was far more scarce, it'd just delete too much stuff and hurt some of the nostalgia aspect (for me at least).

I think what Jiggmin had in mind when he added this was to try and mostly delete accounts that were logged into once or twice and never touched again, which is consistent with most common name accounts (even today, indicating freeing them up doesn't help much anyway). The 50 play count was probably added to specifically make sure very few with published levels were deleted.

I get why you want this, but someone grabbed those names first and were an active player at one point, so you just gotta deal with that unfortunately. At this point in the game's life, I can't really see it helping it more than hurting it anyway. Plus this can lead to impersonation issues; Leader's account was already deleted, and I can count a few other former staff accounts alone that would be affected by this.

bls1999:
For reference, none of my levels have 1000 plays.  I am very very very opposed to making the minimum rank of the user 20 and the play count of levels 1000+.  If anything, I'd say 100 plays would be better and the rank kept at 10.

LP-PR2:
I think i agree with bls on this one, mainly due to the fact that you're just attempting to gather as many alts as you possibly can, this is just a suggestion for your personal gain and nobody elses.

jacob-grahn:
I'm open to bumping it up to 100 plays.

VipaPR2:
Alright @jacob-grahn, sounds good with bumping it up to 100 plays.

jacob-grahn:
@VipaPR2 ok, 15 is fine

VipaPR2:
@jacob-grahn, thanks for merging it, but I also wanted to let you know that it doesn't delete users which have a "userId":null
Here you can see an example: http://pr2hub.com/get_player_info_2.php?name=racer
The account "racer" meets the requirements for deletion, but it doesn't get deleted because it has "userId":null

PR #218, opened 2018-02-25
Add /debug for admins, remove guild de fred from server startup

bls1999:
/debug organizes some useful information for admins that is already public.  Also, show the server loadup time.

When servers restart right now, the Fred body automatically is set as the user's character.  Many players have complained to me and asked why their character is a cactus.  This pull makes sure the perk is supposed to set the part before doing so, also adding `Perks::SERVER_RESTART` to aid with this.

An alternative to this is simply to disable Guild de Fred on server startup, giving you more revenue from the Vault of Magics.

### DIFF
--- a/multiplayer_server/Perks.php
+++ b/multiplayer_server/Perks.php
@@ -6,7 +6,6 @@ class Perks {
 	const GUILD_GHOST = 'guild-ghost';
 	const GUILD_ARTIFACT = 'guild-artifact';
 	const HAPPY_HOUR = 'happy-hour';
-	const SERVER_RESTART = 'server-restart';
 	
 }
 ?>
--- a/multiplayer_server/Perks.php
+++ b/multiplayer_server/Perks.php
@@ -6,6 +6,7 @@ class Perks {
 	const GUILD_GHOST = 'guild-ghost';
 	const GUILD_ARTIFACT = 'guild-artifact';
 	const HAPPY_HOUR = 'happy-hour';
+	const SERVER_RESTART = 'server-restart';
 	
 }
 ?>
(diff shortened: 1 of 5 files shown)


LP-PR2:
I don't think disabling Guild de Fred on startup will give him any revenue, considering nobody even likes the set :man_shrugging:

bls1999:
True.  It may make more sense to disable it on startup just because of that.  Let's take a poll of users on PRG, PR2CD, and JV2.

bls1999:
@jacob-grahn the poll is [here](https://jiggmin2.com/forums/showthread.php?tid=582).  The majority of people who have voted so far seem to either want the Fred parts not to auto-set on the account or just have them disabled completely.

Zelxnte:
I honestly feel the only reason people dislike the Fred set is because of the fact that it's difficult to see where they're standing / it's difficult to glitch in some cases due to the fact that the position where the players' "head" would be is somewhere near the middle after they become Fred (since Fred is bigger/taller) :P 

My guess is most of us are too used to the default settings

jacob-grahn:
Yeah, I'd rather just get rid of fred during hh

bls1999:
Gotcha @jacob-grahn, I can submit a pull for that.

bls1999:
Actually, @jacob-grahn, is there a reason you think it would be better to just disable Fred during HH?

jacob-grahn:
It seems like the solution that requires the least amount of code. :man_technologist:

bls1999:
Haha okay, I can adjust the files accordingly in a bit.

bls1999:
Alright @jacob-grahn, I just removed the server restart perk-related edits I made.

jacob-grahn:
Looks good, thanks!

PR #222, opened 2018-02-26
Make hats be able to fall off in Level Editor

Zelxnte:
I've recently noticed this while making a hat remover of my own...

*It's really a pain to test things like hat removers to see whether they work or not, you've gotta publish the level, wait for it to upload, then refresh newest, and then finally 'test' it.

Instead of this, why not let players decide whether to allow hats to fall off or stay on after hitting mines, etc. That way people who don't want hats to fall off in level editor can keep the option "off" for themselves, while people who want hats to fall off can keep the option "on".

Maybe this can be implemented in with a little toggle button or something of that nature, in the Settings menu of the Level Editor for example:

"-- Removable Hats in Editor --"
[Yes//No]  _or_  [On//Off]
-------------------
--_OR something simpler (very similar to the Item selection setting)_--

"-- Removable Hats in Editor --"
[✔] Enable
----------------------

Maybe keep the default at "Disabled" so people won't get confused :P 

Here's an example image: https://image.prntscr.com/image/ZA4V59tjRjGpVvHSoYU0rw.png
---------------

edit: What do you guys think about this idea?
edit2: Formatted the request

### DIFF
--- a/http_server/cron/delete_old_accounts.php
+++ b/http_server/cron/delete_old_accounts.php
@@ -26,7 +26,7 @@ while($row = $result->fetch_object()) {
 	}
 
 	$str = "$i $user_id plays: $play_count rank: $rank.";
-	if($play_count > 1000 || $rank > 20) {
+	if($play_count >= 100 || $rank >= 15) {
 		output("$str SPARE");
 	}
 	else {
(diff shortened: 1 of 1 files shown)


bls1999:
Since the code for that is hard-coded into the client (and since LE is offline and not on a server), this is currently not possible. :(

Zelxnte:
I guess that means we'll have to wait for the source code?

bls1999:
It it's ever recovered. @zelante

PR #228, opened 2018-02-27
MAJOR security updates, set email exploit fix

bls1999:
This pull is massive, I know.  I'm really sorry about the size... but this is without a doubt the most important pull you will ever receive on this repo.

Let me walk you through some of the changes I made.

First off, you'll see that all of my commits have some kind of abbreviation in them.  Here's what those abbreviations mean:
 - RL: Rate limiting (I put in a lot of this. It will greatly reduce the vulnerability of the database to DDoS attacks)
 - Post: POST request validation
 - Ref: Check the referrer
 - TL: Change `token_login($db)` to `token_login($db, false)`
 - MRES: Change `addslashes()` to `mysqli_real_escape_string()`
 - FN: Change to a different function
 - SC: Add sanity checks

I deleted set_email.php, since 1) it was causing problems (@Myst1cZ's exploit called "ZenTag") and 2) it isn't used by the PR2 client anymore (handled by account_change_email.php).  I also bumped up the guild member limit to 200 at the request of @Jmarroqu.

This pull also contains a fix for a bug with guild emblem file names in guild_info.php, where the file extension would be `.j` instead of `.jpg` by using `substr()` and `str_replace()`.  There's another fix: in admin/update_guild.php, fix owner changes.

I tested all of the scripts with the PR2 client and only changed those that use POST to POST, changed token_login on scripts that pass the token in the URL parameters, etc.  This is not only for your ease of making sure all of this stuff is up-to-par, but also simply because it's good practice to quadruple check your code before it goes in the final version.

This pull is the result of about a week of work.  Please let me know what you think.


VipaPR2:
If it works, good job! :P

bls1999:
@VipaPR2 Well I'm actually deleting the set_email script altogether since it was made useless by a past PR2 update.  So if it doesn't work, then something must have really gone awry...

bls1999:
Fixed a few more things and edited the OP accordingly.

jacob-grahn:
A quick pull seems to be in order! Thanks!

EternalUF:
Holy moly, what a pull. Great job @bls1999 !!!

BrokenBulb78:
Wow, good job!

PR #230, opened 2018-02-28
Fix message reporting again

bls1999:
This is a ballsy move I'm making here, so bear with me.

Instead of using `mysqli_real_escape_string` (which seems to have broken something in checking if the message exists) for the problematic variables, it makes more sense to just convert the strings that should contain numbers (like the user ids, time, etc) to integers.

Theoretically, that should fix it; However, I've been saying that for the past 2 weeks I've been trying to fix this god-awful script.

### DIFF
--- a/http_server/www/message_report.php
+++ b/http_server/www/message_report.php
@@ -4,11 +4,8 @@ header("Content-type: text/plain");
 
 require_once('../fns/all_fns.php');
 
-$message_id = $_POST['message_id'];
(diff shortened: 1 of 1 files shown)


VipaPR2:
Be aware. https://www.youtube.com/watch?v=_jKylhJtPmI

bls1999:
*sigh* fine, I'll work on updating everything to prepared statements...

PR #231, opened 2018-02-28
fix remember me

bls1999:
Call `$_COOKIE['token']` instead of `$_POST['token']`


bls1999:
Combining with a future pull.

PR #233, opened 2018-02-28
 Update register_user.php and guild_create.php

VipaPR2:
When registering an account the list of allowed characters does not match the characters which are actually allowed. I've fixed this and made a slight change to the error message returned when an illegal character is used.


bls1999:
Nicely done!  It may make sense to edit the other places this message appears (guild_edit.php, etc)

jacob-grahn:
Thanks!

PR #236, opened 2018-03-01
Optional: Feet > Foot

Grovbi:
I edit this thing too much, man.

Anyway, changes feet prize pop-ups to instead read "foot" like they did in the olden days. It's _kinda_ misleading since you get more than 1, but the new way has odd grammar (win a: Stick Feet). This could be fixed by adding something to remove "a" from feet prizes like how it changes to "an" for vowels, but (I think) that'd require the client, sadly.


jacob-grahn:
:man_shrugging: OK!

PR #240, opened 2018-03-01
Fixes, PDO

bls1999:
I figured might as well get started on PDO conversion with the syntax of SQL queries I know.  This one converts both bans.php and ban_user.php to PDO.

The ban queries (see queries/staff/ban_user.php) and the user info query should all work fine (see queries/get_info/user.php); However, please do check my action log function syntax (see queries/staff/action_log.php).

Your to-do:
 - Figure out a way to log if there's more than 1 user found for an ID or name
 - Check my PDO syntax in queries/staff/action_log.php
 - Make the current stored procedures public

My to-do:
 - See what other plain text queries I can convert
 - Convert the queries when you make them public


jacob-grahn:
I'm not super familiar with the pdo syntax yet, I'll have to study up.

Is the `throttle_bans` function still needed with the rate limiting you added?

bls1999:
Yeah, throttle_bans is still needed since trial/perma mods have a different number of maximum bans per hour.

Also, the ban list is currently broken.  Can you change the file so that the SQL query is enclosed in double quotes instead of single ones?

bls1999:
Also, here's a walkthrough of the `user_select_by_name` function:

```
function user_select_by_name($pdo, $name) {
	$stmt = $pdo->prepare('SELECT * FROM users WHERE name = ?'); // prepare the query
	$stmt->bindValue(1, $name, PDO::PARAM_STR); // bind the value (1 specifies that this variable replaces the first ?, $name is the name passed to the string, PDO::PARAM_STR specifies that it's a string
	$stmt->execute(); // do it
	$user = $stmt->fetchAll(PDO::FETCH_OBJ); // get the results as an object (would be $result->fetch_object() in regular php)
	return $user; // returns the data
}
```

Line-by-line:

aaaaaa123456789:
As I told you on Discord a while ago, testing `$stmt -> rowCount() == 1` before fetching the row might be some helpful error-checking, which we all know this codebase wholly lacks.

http_server/queries/get_info/user.php
```diff
@@ -0,0 +1,29 @@
...
+function user_select_by_name($pdo, $name) {
+	$stmt = $pdo->prepare('SELECT * FROM users WHERE name = ?');
+	$stmt->bindValue(1, $name, PDO::PARAM_STR);
+	$stmt->execute();
+	$row_count = $stmt->rowCount();
+	if ($row_count < 1) {
+		return false;
+	}
+	// TODO: Add check for more than 1 result and log critical failure if so.
+	$user = $stmt->fetchAll(PDO::FETCH_OBJ);
```
jacob-grahn: Does fetchAll return an array? It looks like fetchObject will work: http://php.net/manual/en/pdostatement.fetchobject.php

Or! Since there doesn't seem to be a better way to get the row count, we could do
```
$stmt->execute();
$results = $stmt->fetchAll();
if (count($results) !== 1) {
  // throw an error
}
$user = $results[0]; // I believe this defaults to being an associative array
return $user;

File: http_server/queries/get_info/user.php [hunk 2]
```diff
@@ -0,0 +1,29 @@
...
+function user_select_by_id($pdo, $user_id) {
+	$stmt = $pdo->prepare('SELECT * FROM users WHERE user_id = ?');
+	$stmt->bindValue(1, $user_id, PDO::PARAM_INT);
+	$stmt->execute();
+	$row_count = $stmt->rowCount();
+	if ($row_count < 1) {
+		return false;
+	}
+	// TODO: Add check for more than 1 result and log critical failure if so.
+	$user = $stmt->fetchAll(PDO::FETCH_OBJ);
```
jacob-grahn: Same comments as user_select_by_name

http_server/queries/staff/ban_user.php
```diff
@@ -0,0 +1,39 @@
...
+	$stmt->bindValue(3, $mod_user_id, PDO::PARAM_INT);
+	$stmt->bindValue(4, $time, PDO::PARAM_INT);
+	$stmt->bindValue(5, $expire_time, PDO::PARAM_INT);
+	$stmt->bindValue(6, $reason, PDO::PARAM_STR);
+	$stmt->bindValue(7, $record, PDO::PARAM_STR);
+	$stmt->bindValue(8, $banned_name, PDO::PARAM_STR);
+	$stmt->bindValue(9, $mod_name, PDO::PARAM_STR);
+	$stmt->bindValue(10, $ip_ban, PDO::PARAM_INT);
+	$stmt->bindValue(11, $account_ban, PDO::PARAM_INT);
+
```
jacob-grahn: What if we use named parameters if there are more than a few question mark placeholders? It seems like it that would help prevent leaving something out or getting the order wrong.

```
VALUES (:banned_id, :banned_user_id, ...);
$stmt->bindValue(':banned_id', $banned_ip, PDO::PARAM_STR);
$stmt->bindValue(':banned_user_id', $banned_user_id, PDO::PARAM_INT);
```

http_server/queries/get_info/user.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+function user_select_by_name($pdo, $name) {
+	$stmt = $pdo->prepare('SELECT * FROM users WHERE name = ?');
+	$stmt->bindValue(1, $name, PDO::PARAM_STR);
+	$stmt->execute();
+	$row_count = $stmt->rowCount();
```
jacob-grahn: It looks like `rowCount()` doesn't work with `SELECT` queries. 

http://php.net/manual/en/pdostatement.rowcount.php

http_server/queries/get_info/user.php
bls1999: @aaaaaa123456789 would this work?

http_server/queries/staff/ban_user.php
bls1999: Yeah I can do that, let me edit this real quick.

http_server/queries/get_info/user.php
bls1999: What if we did something like this?
```
$stmt = $pdo->prepare('SELECT *, COUNT(*) as row_count FROM users WHERE name = ?');
$stmt->bindValue(1, $name, PDO::PARAM_STR);
$stmt->execute();
$result = $stmt->fetchAll(PDO::FETCH_OBJ);
if ($result->row_count < 1) {
   // etc...
}
```

bls1999: rowCount() works fine on my localhost with a SELECT query, so I'm not sure why php.net says it doesn't work :thinking:

http_server/queries/get_info/user.php
bls1999: @jacob-grahn the problem with using fetchAll by itself (at least that I've encountered) is that the columns are duplicated (instead of 6 array keys/values, we have 12).  Using fetchObject only gets one result, which _theoretically_ works for this scenario, although I think it would be cleaner to just use `fetchAll(PDO::FETCH_OBJ)` since there isn't really a downside.

To get an associative array, you'd replace `PDO::FETCH_OBJ` with `PDO::FETCH_ASSOC` in `fetchAll()`.

bls1999: @aaaaaa123456789 please correct me if I'm wrong, here.

@jacob-grahn the problem with using fetchAll by itself (at least that I've encountered) is that the columns are duplicated (instead of 6 array keys/values, we have 12).  Using fetchObject only gets one result, which _theoretically_ works for this scenario, although I think it would be cleaner to just use `fetchAll(PDO::FETCH_OBJ)` since there isn't really a downside.

To get an associative array, you'd replace `PDO::FETCH_OBJ` with `PDO::FETCH_ASSOC` in `fetchAll()`.

http_server/queries/staff/ban_user.php
bls1999: done!

PR #242, opened 2018-03-02
Add /dc, /kick, rework code in Player.php

bls1999:
Adds the /dc and /kick commands for trial/permanent moderators.  Also fixes some code dynamics.


jacob-grahn:
Looks good!

PR #243, opened 2018-03-02
Update delete_old_accounts.php

VipaPR2:
Not entirely sure if this is a viable method to fix "userId: null" users not being deleted even though they meet the requirements for deletion however I'm fairly positive it should work. I also wanted to inquire as to how often this script is executed, would it be possible for you to re-run the script after adding these fixes?

### DIFF
--- a/http_server/cron/delete_old_accounts.php
+++ b/http_server/cron/delete_old_accounts.php
@@ -24,6 +24,14 @@ while($row = $result->fetch_object()) {
 	if(!isset($play_count)) {
 		$play_count = 0;
 	}
+	if(!isset($user_id)) {
+    		$user_id = 0; //placeholder id so it isn't null
+    		output("$str DELETE");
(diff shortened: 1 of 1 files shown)


VipaPR2:
Any idea if it's possible to make it work in any way?

bls1999:
I'd need to know the queries sent to the database

VipaPR2:
Alright, Jiggmin will see this thread and he can maybe fix it.

bls1999:
The issue is that the pr2_insert procedure isn't called when a user registers, it's called when a user logs in for the first time.  I think in order for the script to determine user ID, it goes through pr2_select instead of user_select.  If this is the case, it could be theoretically be fixed by going through user_select instead.

jacob-grahn:
@VipaPR2 could you open this as an issue? I think the solution @bls1999 suggested would probably work.

PR #244, opened 2018-03-02
Fixes/additions to http_server and multiplayer_server

bls1999:
This pull is a bunch of small updates that have combined to form one normal-sized pull.

HTTP Server Fixes/Additions:
 - A possible exploit where users could possibly log in as other users.  Fire-Bird was able to demonstrate something like this and I hope this fixes it (if it's not already fixed).
 - An error where accounts missing pr2 data would break update_account.php
 - An error where changing users' guild IDs in update_account.php would not increment the guild member count
 - A parenthesis error in the ban log that lets people make huge queries to the database.
 - An error that broke player_search.php
 - A small fix in search_by_email.php
 - HOPEFULLY fix F@H awards for good (the issue is that new folders aren't being made accounts in the fah database. I'm hoping that 53acfea will fix it, but PLEASE check my work to make sure).
 - Add `pr2_insert` into register_user.php to avoid broken accounts

Multiplayer Server Fixes/Additions:
 - Fix /kick (and add mod action logging for the kick function)
 - Fix /dc (and add mod action logging for the command)
 - Add test function `become_server_owner()` (right now just for Fred)
 - Add `rate_limit`, `name_to_id`, and `id_to_name` to fns/utils.php
 - Add rate limiting to recording play counts
 - Add EXP hat nerfing on Campaign (see issue #245)
 - Add player argument to admins' /debug command

Please check to make sure my port of the rate limiting function will work properly.  I'm not familiar with how the apcu functions work, so it very well could just not be compatible.

### DIFF
--- a/http_server/cron/fah_award_prizes.php
+++ b/http_server/cron/fah_award_prizes.php
@@ -2,7 +2,6 @@
 
 require_once(__DIR__ . '/../fns/all_fns.php');
 
-
 $prize_array = array();
 $processed_names = array();
 $name_switch_array = array();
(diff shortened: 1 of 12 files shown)


Myst1cZ:
The numbers that cause the huge database query:

jacob-grahn:
\*normal-sized pull\*
*12 files changed*
*430 lines added*
*257 lines removed*
:roller_coaster:

bls1999:
LOL @jacob-grahn

PR #252, opened 2018-03-06
One more fix in player_search.php

BrokenBulb78:
It was still throwing an error when you were loading the page, it should be fixed now I guess.


jacob-grahn:
I don't think that'll work, but I do like your other formatting changes. I'll merge this in and work on it a bit.

PR #254, opened 2018-03-06
Fix server crash, change artifact bonus

bls1999:
1 dot. Oy vey.

Also, the artifact bonus is now a maximum of 50k, but the rank where it reaches 50k is rank 60 instead of 20.


BrokenBulb78:
That dot made the server crash :joy:

PR #255, opened 2018-03-06
Slight fixes to moderation.php and promote_to_moderator.php

bls1999:


jacob-grahn:
Database calls in the multiplayer server will force the game loop to wait until the query is completed. The queries set up with MYSQLI_ASYNC get around the problem by just not waiting for an answer. Things like `name_to_id` need an answer though, so it would be better to avoid using it if at all possible.

bls1999:
@jacob-grahn we'd have to make `name_to_player` case-insensitive because right now it's case-sensitive.

PR #261, opened 2018-03-07
Botting patch, slight fixes

bls1999:
HTTP Server:
 - Fixed query in queries/get_info/user.php
 - Created query for pr2_select (check the code, commit 3e97cd5)
 - Fixed player_info.php and semi-updated it to PDO
 - Fixed server crash on admin server owner logout
 - Fixed leaderboard.php

Socket Server:
 - Made `name_to_player()` case insensitive and removed `name_to_id()` calls
 - Fixed/optimized moderation.php, promote_to_moderator.php
 - Disabled and removed +5 EXP bonus
 - Added /hh command (see comments on 93c168a for more details) 
 - Added a check for if the user is winning the artifact (see comments on 927f7f9). Hopefully it works?

There have been reports of botting on levels with a set prize with the +5 EXP bonus, and it's confirmed by those levels' play counts (see The Golden Compass and Hat Factory as examples).  Basically, they enter a level and auto-win as soon as the race starts, collecting the +5 EXP.  Do this a bunch of times and you get an exorbitant amount of EXP.

The easy solution here is to remove the +5 EXP bonus because:
 - The bonus popup doesn't show up in the client anymore
 - It has no effect since EXP has been inflated over the years
 - It would be a lot easier than trying to patch the auto-win glitch, which may require a client edit

If we can patch the auto-win glitch, I wouldn't be opposed to putting the bonus back in the future; However, right now this is a quick fix to stop botting.

### DIFF
--- a/http_server/queries/get_info/pr2.php
+++ b/http_server/queries/get_info/pr2.php
@@ -1,0 +1,22 @@
+<?php
+
+function pr2_select_by_id($pdo, $id)
+{
+	$stmt = $pdo->prepare('SELECT * FROM pr2 WHERE user_id = :id');
+	$stmt->bindValue(':id', $id, PDO::PARAM_INT);
+	$stmt->execute();
+	
+	// get number of results
+	$row_count = $stmt->rowCount();
+	if ($row_count < 1) {
+		return false;
+	}
+	if ($row_count > 1) {
+		// log critical failure
+		return false;
+	}
+	
+	// return result
+	$result = $stmt->fetchAll(PDO::FETCH_OBJ);
+	return $result;
+}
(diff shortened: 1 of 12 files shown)


jacob-grahn:
Looks good, I'll get this running on fred

EternalUF:
Woooo! This pull pushed the repo past 1,000 commits congrats all!!!

LP-PR2:
Yay lets not earn any extra exp for winning parts :) way to ruin it for the rest of us.

bls1999:
You can thank the botters for that @LP-PR2.

LP-PR2:
What exactly are they botting and how is it beneficial? Like do you have any proof?

LP-PR2:
I'm sure you do considering who you're getting your information from, i just think it's petty to remove something so small.

PR #262, opened 2018-03-08
Add debugging to /hh (figuring out the problem)

bls1999:
Also, disable it from being used by members for the time being so it doesn't crash the server.


bls1999:
Additionally, @jacob-grahn, can you post the error/crash log for Derron for the past few hours?  It would make my life a lot easier. :)

jacob-grahn:
It looks like we've got this working now

PR #264, opened 2018-03-09
Beautified bans.php

BrokenBulb78:
I beautified Ban Log a bit.

image


bls1999:
Hey Anton, as someone who uses that list all the time, I can tell you that would not be a welcome change.

LP-PR2:
That looks hideous

PR #265, opened 2018-03-09
Hopefully fix server crash on already having prize

bls1999:
Apparently the server crashes if you already have the prize that's being offered.  I'm hoping this fixes it.

Also, fix a variable in leaderboard.php and fix the time display in `HappyHour::timeLeft()`


jacob-grahn:
Not sure if that'll fix it. Maybe something is still looking for the +5exp prize and failing to find it?

bls1999:
Not sure. I thought I removed all instances of the +5 exp bonus from the client…

PR #271, opened 2018-03-11
Update Game.php

Light337:


bls1999:
Just wondering: what's the point of your change on line 53?

Light337:
Currently, the statement is preventing some people from joining races even when there's a blank/empty slot. This should resolve the issue for now :)

Light337:
Both, actually. Also, this should fix the issue where you're kicked out of a race enter zone just after clicking on the race.

Light337:
@jacob-grahn

EternalUF:
I’m unsure if I’ve ever ran into this issue before...

Light337:
What botting? This is supposed to fix a problem, not cause a problem. Have you never had an issue where you join a race and then you're kicked out of the waitlist without clicking anything? I could probably reproduce the error quickly.

bls1999:
Could you make a video?

Light337:
Sure, give me a sec.

Light337:
Have to upload it to YT. Few minutes.

Light337:
https://www.youtube.com/watch?v=ENv0x2qSbQk&feature=youtu.be

Easiest way to reproduce the error, but this can also happen at times when trying to play a normal race with others or even when playing solo races.

bls1999:
It isn't apparent as to what the actual bug or issue is in that video, but from what I can tell, I believe you're triggering the loiter detector (which isn't a bug).  Even if it was, it wouldn't be changed in Game.php, as that php file handles what happens after you move from the lobby to the race.

@jacob-grahn

jacob-grahn:
It seems like this would allow infinite players into a game?

LP-PR2:
Rip @light1337

aaaaaa123456789:
Someone linked me to this lulfest of a commit. This would indeed enable botting (kinda like the Firebird bot that took accounts to rank 60 back in 2012) and at the same time only grant prizes for races with three or more players (as opposed to two or more as the engine stands).

This is an intentionally harmful pull request made with the clear intent of sneaking in an exploit into the server. I'd recommend banning this account from making any further pull requests, and automatically rejecting all pulls that either have no description or have a description that doesn't match their contents.

EternalUF:
Although I agree with this 10000000% keep in mind this GitHub Account was made the same day the pull was requested, so obviously an alt/troll account. I see this as just a friendly reminder to CAREFULLY inspect all pull requests.

PR #282, opened 2018-03-16
Modify/add PDO queries, update files

bls1999:
Add error reporting to some, optimize queries.


jacob-grahn:
:+1: Looks good, thanks!

PR #284, opened 2018-03-19
Fixing Player Overflow

Paladin21:
This is a fix that prevents slots from being refilled. Refilling slots can cause a player overflow in races. The only time a slot should be refilled is when it doesn't have a numerical value.


bls1999:
This is another phony pull request made by an account that joined very very recently. @jacob-grahn 

This would not prevent anything.  It's fine as it is right now.

aaaaaa123456789:
Anyone with minimal knowledge of programming can tell that `($slot < 0) && ($slot > 3)` is bogus code. This is a malicious patch; essentially a repeat of #271.

PR #289, opened 2018-03-20
Fixes to admin tools, PDO queries

bls1999:
I've been finding a few queries with `p_` before some of the variables.  I assume this has to do with stored procedures?  I've been converting them as I find them to PDO values.

Also, if you could post the error log for /admin I'd be a very happy man.

Finally, what is the query for `user_select_expanded`?


jacob-grahn:
Yeah, the `p_`s are a leftover from the stored procedures. Looks like I've begun to loose my mind and leave them in there.

`user_select_expanded` was in the last batch of updates: https://github.com/jacob-grahn/platform-racing-2-server/blob/master/http_server/queries/users/user_select_expanded.php

PR #290, opened 2018-03-20
Fix level search

bls1999:
Someone was feeling a little reckless and switched level search by title/user 😂 @jacob-grahn


jacob-grahn:
Nice :+1:

PR #291, opened 2018-03-20
Continue the PDOcalypse, fix a few things

bls1999:
Fixes:
 - No change validation -> update_account.php
 - Check for a non-existent guild when changing user into a new guild -> update_account.php
 - Fix IP history -> player_deep_info.php
 - Fix user_select_expanded
 - Fix bans
 - Fix level unpublishing (check my commit comments on cf8a7e7)
 - Fix level searches (they were only searching by user)
 - Revert curly braces
 - Rearrange some functions and calls
 - Fix random things broken in mod/player_info.php
 - Display "none" on no guild in player search

Nice things:
 - Admins now have some friendly links on guild_deep_info and player_deep_info

It sounds like a lot, but most of it was small changes in a few files.  One thing to note going forward: curly braces don't work the way you used them in 851c7bc, dadd592, and 7de6073.

Lastly, for the life of me, I cannot figure out why the heck `changing_email_insert` isn't working as it should.  I'm assuming the table `changing_emails` is a copy of the table `guild_transfers`, so it's seriously stumping me as to why it's not working.  See 891d99d for more information.

### DIFF
--- a/http_server/queries/changing_emails/changing_email_insert.php
+++ b/http_server/queries/changing_emails/changing_email_insert.php
@@ -4,23 +4,23 @@ function changing_email_insert($pdo, $user_id, $old_email, $new_email, $code, $i
 {
     $stmt = $pdo->prepare('
         INSERT INTO changing_emails
-        SET user_id = :user_id,
-            old_email = :old_email,
-            new_email = :new_email,
-            code = :code,
-            request_ip = :ip,
-            status = "pending",
-            date = NOW()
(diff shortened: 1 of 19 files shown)


jacob-grahn:
Hmm, I was trying to use the complex string syntax [here](http://php.net/manual/en/language.types.string.php#language.types.string.parsing), but it seems it was too complex indeed.

I'll take a look at `changing_email_insert`...

PR #297, opened 2018-03-22
Fix player_deep_logins.php, messages_get.php

bls1999:
Looks like the issue with the messages was it was pushing the raw database result as the return data of messages_get.php.  Hopefully changing the variable will fix this (foreach may be doing weird things to the variable scope).


jacob-grahn:
Awesome, thanks!

PR #298, opened 2018-03-22
Add RL to admin tools, misc fixes/optimizations

bls1999:
- Adds rate limiting, extra checks/validations to admin tools
 - Continuity (assoc -> obj, coding style, etc)
 - Remove `name_to_id` and `id_to_name` functions in query_fns.php but require them in that script.
 - Edit error messages
 - Fix daily server restarts
 - Fix guild member incrementing
 - Fix user registration
 - Fix leaderboard finding only rank 50s without tokens


http_server/queries/users/users_select_top.php
```diff
@@ -10,7 +10,7 @@ function users_select_top($pdo, $start, $count)
         FROM users, pr2, rank_tokens
         WHERE users.user_id = pr2.user_id
         AND pr2.user_id = rank_tokens.user_id
-        AND rank > 49
+        AND active_rank > 49
```
jacob-grahn: I tried this same thing. It seems a computed value can't be used in this part of the query. :/

jacob-grahn:
Many thanks! :+1:

http_server/queries/users/users_select_top.php
bls1999: @jacob-grahn could `HAVING` be used instead? https://forums.mysql.com/read.php?10,266846,267031#msg-267031

PR #299, opened 2018-03-22
Make server owners admins on their server, add become_guest()

bls1999:
- Server owners are now server admins and can promote temp mods on their server
 - Testing the new `become_guest()` function I added to Player.php for badges support on Kong (only one account is enabled to use this function on login)
 - Make HH on server restart 30 mins and a 1/10 chance of occurring
 - Add /rules command and change the welcome message for main chat to include the rules link

### DIFF
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -654,6 +654,13 @@ class Player
                 } else {
                     $this->write("Invalid action. For more information on how to use this command, type /mod help.");
                 }
+            } 
+            elseif ($chat_message == '/rules') {
+                $message = 'systemChat`The PR2 rules can be found here: https://jiggmin2.com/forums/showthread.php?tid=385.';
+                if ($guild_id != 0) {
+                    $message .= ' Since this is a private server, your guild owner may have different rules for the chatrooms and the server. Check with them if you\'re unsure.';
+                }
+                $this->write($message);
             } // help command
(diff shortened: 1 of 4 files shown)


LP-PR2:
"Make HH on server restart 30 mins and a 1/10 chance of occurring"

When were the HHs an issue?

fourst4r:
> Make HH on server restart 30 mins and a 1/10 chance of occurring

...why?

I always thought of the happy _hour_ during server restart as like a consolation for the inconvenience of having your game interrupted.

Now you propose we have a 1 in 10 chance to get _half_ of an HH? This is an over-complicated fix for an issue that doesn't exist.

LP-PR2:
"Theoretically, one server will get a 30-min happy hour at the daily server restarts"

That's not a happy hour, and what if that one server is a private server? Because those exist too.

bls1999:
The idea is that instead of all of the servers getting spammed with full Happy Hours once a day, the occurrence is random but the chance is way higher.  I would be fine with making the time 60 minutes again, although the thought there was to have sort of a mini Happy Hour for a server restart instead of a full one.

LP-PR2:
@bls1999 why not make it a certain hh on 1 server? That way it's not spamming all the servers and it's guaranteed to be on a public server and not a guild server.

bls1999:
Sounds good to me.

bls1999:
I put it on Carina ^^

PR #305, opened 2018-03-23
Revert queries back to pre-#298

bls1999:
Reverting the `users_select_hash_by_name` query to what it was before my changes, then changing `user_select` in `query_fns.php` back to `users_select_hash_by_name`.


bls1999:
Wrong place lol

PR #308, opened 2018-03-24
Fixes to admin tools, queries, www files

bls1999:
- Make ban_edit.php way less convoluted
 - Cosmetic edits to player_deep_logins.php
 - Restructure some error messages so they're not as ugly to regular users
 - Add support for getting player info by user id in get_player_info_2.php
 - Attempt to fix guild members not incrementing correctly with admin tools
 - Add a kong link to the list of approved sites
 - Hopefully fix an error in recent_logins that causes logins not to record
 - Fixed being able to rate your own levels
 - [Hopefully] fixed passwords on all new levels


jacob-grahn:
:+1:

PR #314, opened 2018-03-26
Fix action logging, ban_edit.php, set_campaign.php

bls1999:
Hopefully*


jacob-grahn:
Hmmm, interesting that `{}` would be problematic in a double quoted string.

PR #315, opened 2018-03-26
More tests for become_guest()

bls1999:
It seems that logging into a guest account with a password (email me if you're interested how I do this) yields different results than clicking "Play As Guest".  I need to perform some more tests.


jacob-grahn:
:+1:

PR #316, opened 2018-03-27
Format queries, error messages

bls1999:
I haven't done a humungous pull in a while... let's change that!

This pull goes through all of the queries and formats them neatly.  It also aims to standardize the formatting and error reporting of each file.

I've used the following keywords in most of the commits:
 - err msg: means I edited the error message (I used this interchangably with format/formatting)
 - formatting/format: means I edited the code structure
 - is it/this needed: is it needed?

Also, these files have queries that I don't know how to separate:
 - [queries/artifact/artifact_check.php](https://github.com/jacob-grahn/platform-racing-2-server/blob/master/http_server/queries/artifact/artifact_check.php)
 - [queries/fah/stats/stats_insert.php](https://github.com/jacob-grahn/platform-racing-2-server/blob/master/http_server/queries/fah/stats/stats_insert.php)

Let me know if I can provide any more info!

### DIFF
--- a/http_server/queries/bans/ban_lift.php
+++ b/http_server/queries/bans/ban_lift.php
@@ -13,9 +13,10 @@ function ban_lift($pdo, $ban_id, $lifted_by, $lifted_reason)
     $stmt->bindValue(':ban_id', $ban_id, PDO::PARAM_INT);
     $stmt->bindValue(':lifted_by', $lifted_by, PDO::PARAM_STR);
     $stmt->bindValue(':lifted_reason', $lifted_reason, PDO::PARAM_STR);
-
     $result = $stmt->execute();
+    
     if ($result === false) {
+        $ban_id = (int) $ban_id;
         throw new Exception("Could not lift ban #$ban_id.");
     }
 
--- a/http_server/queries/messages_reported/messages_reported_check_existing.php
+++ b/http_server/queries/messages_reported/messages_reported_check_existing.php
@@ -16,9 +16,5 @@ function messages_reported_check_existing($pdo, $message_id)
     
     $count = $stmt->fetchColumn();
     
-    if ((int) $count == 0 || $count == false || empty($count)) {
-        return false;
-    }
-    
-    return true;
+    return (bool) $count;
 }
--- a/http_server/queries/messages_reported/messages_reported_check_existing.php
+++ b/http_server/queries/messages_reported/messages_reported_check_existing.php
@@ -14,7 +14,5 @@ function messages_reported_check_existing($pdo, $message_id)
         throw new Exception("Could not check if this message has been reported already.");
     }
     
-    $count = $stmt->fetchColumn();
-    
-    return (bool) $count;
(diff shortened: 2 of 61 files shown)


http_server/queries/bans/ban_lift.php
```diff
@@ -16,7 +16,7 @@ function ban_lift($pdo, $ban_id, $lifted_by, $lifted_reason)
 
     $result = $stmt->execute();
     if ($result === false) {
-        throw new Exception('Could not lift ban');
+        throw new Exception("Could not lift ban #$ban_id.");
```
jacob-grahn: If this is echoing out to an html page, it possibly needs htmlspecialchars?

http_server/queries/messages_reported/messages_reported_check_existing.php
```diff
@@ -16,7 +16,7 @@ function messages_reported_check_existing($pdo, $message_id)
     
     $count = $stmt->fetchColumn();
     
-    if ($count == 0) {
+    if ((int) $count == 0 || $count == false || empty($count)) {
```
jacob-grahn: It seem like `return (boolean) $stmt->fetchColumn();` would take care of all of the possibilities here.

Also, cool trick with the fetchColumn. Didn't know that one existed :)

http_server/queries/servers/servers_select_highest_port.php
```diff
@@ -2,20 +2,19 @@
...
 {
-    $stmt = $pdo->prepare('
-        SELECT servers.port
-        FROM servers
-        ORDER BY servers.port DESC
-        LIMIT 1
+    $result = $pdo->exec('
+        SELECT port
+          FROM servers
+         LIMIT 1
```
jacob-grahn: without an `ORDER BY port DESC` line, this may not return the highest port #

http_server/queries/users/user_delete.php
```diff
@@ -1,91 +1,401 @@
 <?php
```
jacob-grahn: Dang, this must have taken forever :+1:

http_server/queries/bans/ban_lift.php
bls1999: I can add an integer type conversion.

http_server/queries/messages_reported/messages_reported_check_existing.php
bls1999: Aha! Neat. Learn something new every day, I guess. :)

http_server/queries/servers/servers_select_highest_port.php
bls1999: u rite, I'll fix that

http_server/queries/users/user_delete.php
bls1999: Copy/paste is the best thing ever known to man.  Still, it took about 20 minutes just for this file. :expressionless:

jacob-grahn:
:+1:

PR #318, opened 2018-03-27
Prevent forcing log out

Myst1cZ:
Fix to prevent automatic log out through things such as image or web page.

### DIFF
--- a/http_server/www/logout.php
+++ b/http_server/www/logout.php
@@ -14,6 +14,10 @@ try {
 
     if (isset($_COOKIE['token'])) {
         // connect to the db
+        $ref_address = check_ref();
+        if ($ref_address !== true){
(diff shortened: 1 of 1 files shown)


bls1999:
The reason I didn't do this is because some clients are embedded on different sites other than what is allowed.  It doesn't make sense not to allow logouts from everywhere.

bls1999:
Also, it's not a big deal if you get logged out.  That can't cause any lasting harm; it's only a mild annoyance.

jacob-grahn:
Yeah, I don't think this one would work

jacob-grahn:
What if we store the referrer in the cookie when it's created, and then ignore the cookie if it gets sent from a different referrer?

Myst1cZ:
@jacob-grahn Not sure if that would be enough since PHP setcookie or javascript document.cookie could be used to set the referrer in cookie before sending request.

PR #321, opened 2018-03-28
Fix action logging, expire old email change/guild transfer requests

bls1999:
This update fixes action logging.  It also expires email change and guild transfer requests older than a day for security purposes.  Feel free to change the amount of time on that if you'd like, but I think that the least amount of time the changes are valid, the bigger a win for security.


jacob-grahn:
:+1:

PR #322, opened 2018-03-28
Hopefully fix guests earning Kong badges

bls1999:
I think I determined the reason why guest accounts don't earn Kong badges... and it's a weird one.

Apparently, if the pass_login function isn't used, the game just won't let badges be earned.  I have no idea why this is the case, but I've confirmed it by using the guest password to log into a specific guest account and comparing the result with both typing "Guest" in the login popup and clicking the "Play As Guest" button on the main menu.  The only other possibility (which I'm hoping isn't the case) is the game checks for a password value being entered client-side.  If that's the case, I'm afraid we're out of luck unless we find a way to hack the client.

I can LYK more about how I logged into a specific guest account privately.


jacob-grahn:
Can normal accounts earn badges if `pass_login` isn't used?

jacob-grahn:
Also, can you send me the value for $GUEST_PASS?

PR #323, opened 2018-03-28
[Tiny pull] Add host_ip to contest_winner_insert

bls1999:
Forgot one thing.  Can you add this column to the db?  `varchar(15)` should suffice.


jacob-grahn:
Will do, thanks!

PR #333, opened 2018-03-29
Three small fixes

bls1999:
In keeping with the theme of this pull (#333), there are three commits and three files changed in this pull:
 - Fix changing usernames to a different case and guild incrementing in manual guild changes in update_account
 - Don't let Fred enter the admin room
 - Fix a broken variable in Player.php

### DIFF
--- a/http_server/www/admin/update_account.php
+++ b/http_server/www/admin/update_account.php
@@ -167,7 +167,7 @@ function update($pdo, $admin)
     }
     
     // make sure the name doesn't exist
-    if ($user->name != $user_name) {
+    if (strtolower($user->name) != strtolower($user_name)) {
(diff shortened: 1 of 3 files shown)


jacob-grahn:
How is fred able to enter the admin chat in the first place? Is it just on Fred's server?

EternalUF:
@jacob-grahn It's because Fred is the server owner of all public servers, and bls1999 gave more powers to server owners, now server admins as seen in #299 So, Fred is technically the server admin of the public servers, but the only powers Fred has is the ability to make temp mods, which should be restricted soon if not already and being able to join admin channel, which is being restricted here. Hopefully that all grammatically made sense and answered your question...

jacob-grahn:
Thanks @EternalUF

PR #335, opened 2018-03-30
Misc fixes

bls1999:
HTTP Server:
 - Add `is_admin()` function (the same thing as is_moderator but for admins)
 - Fix message reporting
 - [Hopefully] fix ban editing
 - [Hopefully] show users who don't have an entry in `rank_tokens` on the leaderboard
 - Some XSS prevention
 - Use new queries in #347 and #348 with `send_pm()` in `pr2_fns.php`:
    - `pr2_select_true_rank` to fix issue #336
    - `user_select_power` to avoid getting unneeded data
 - Fix a bug that reverts usernames to the registered account case (i.e. logging in as BLS would become bls if registered as bls)

Socket Server:
 - Fix an error where microseconds more than .9950 and less than 1.0000 would display as .100 on the client.  Shout out to @aaaaaa123456789 for simplifying my code here.
 - Fix Fred being able to promote temps


aaaaaa123456789:
And yet you didn't fix the spelling of "obscene"...

multiplayer_server/rooms/Game.php
```diff
@@ -427,12 +427,20 @@ public function remote_finish_race($player, $data)
         }
         $this->finish_race($player);
     }
+    
+    
+    private function ensure_time_format($time) {
+        if ($time < 0) {
+            $time = 0;
+        }
+        return sprintf("%.2f", $time);
```
aaaaaa123456789: Since you don't actually _want_ a string, `return round($time, 2);` would be a better choice.

bls1999:
I can't even use being tired as an excuse for that one.  I committed that hours ago!  Just fixed it, because if I didn't, that would've really bothered me. 🤷‍♂️

http_server/fns/data_fns.php
```diff
@@ -319,7 +319,23 @@ function is_moderator($pdo, $check_ref = true)
...
 
-//
+// returns true if you are logged in as an admin, false if you are not
+function is_admin($pdo, $check_ref = true)
+{
+    $is_admin = false;
+    try {
+        check_moderator($pdo, $check_ref, 3);
+        $is_admin = true;
+    } catch (Exception $e) {
```
aaaaaa123456789: Bad idea to have completely empty catch blocks. But if you really aren't doing anything with the exception, don't declare it; `catch {}` is perfectly valid.

http_server/fns/data_fns.php
```diff
@@ -249,7 +249,7 @@ function talk_to_server($address, $port, $key, $server_function, $receive = fals
 
 
 
-//--- tests to see if a string contains obsene words ---------------------------------------
+// tests to see if a string contains obscene words
 function is_obsene($str)
```
aaaaaa123456789: Fix the spelling of the function too!

multiplayer_server/rooms/Game.php
bls1999: What about typecasting to a float?

aaaaaa123456789: It already is a float. What you want is to have two decimal places exactly, so you can't have .991 or .997 or any other problematic value. That's what `round()` does.

http_server/fns/data_fns.php
bls1999: I just got in bed, so I think I'll wait until tomorrow to figure out everywhere that function is used 😂

multiplayer_server/rooms/Game.php
bls1999: Gotcha.  If I don't add this first, @jacob-grahn will.

bls1999:
fixed the "obscene" typo, ready to merge

jacob-grahn:
:+1:

PR #344, opened 2018-03-30
fix double ip check

fourst4r:
don't need this since it's checked in pass_login


fourst4r:
This would not allow banned IPs to log in, a guest login is now identical to a user login. 
This means we can let `pass_login` handle the ban checking.

bls1999:
Makes sense! Nice catch.

jacob-grahn:
:+1:

PR #345, opened 2018-04-01
Error Handling and Code Cleanup

MoonWithSunshine:
Start_time is always given a value now to prevent subtracting by a null value (which aint good). I also decided to tinker with the probability of receiving hats and made it slightly easier in the long run for newer players.

### DIFF
--- a/multiplayer_server/rooms/Game.php
+++ b/multiplayer_server/rooms/Game.php
@@ -15,7 +15,7 @@ class Game extends Room
 
     private $finish_array = array();
     private $course_id;
-    private $start_time = 0;
+    private $start_time;
     private $begun = false;
     private $loose_hat_array = array();
     private $next_hat_id = 0;
--- a/multiplayer_server/rooms/Game.php
+++ b/multiplayer_server/rooms/Game.php
@@ -15,7 +15,7 @@ class Game extends Room
 
     private $finish_array = array();
     private $course_id;
-    private $start_time;
+    private $start_time = 0;
     private $begun = false;
     private $loose_hat_array = array();
     private $next_hat_id = 0;
@@ -234,17 +234,17 @@ class Game extends Room
             }
         }
 
-        if (!isset($this->prize) && $player_count >= 2) {
-            if (rand(0, 40) == 40) {
+        if (!isset($this->prize) && $player_count >= 1) {
+            if (rand(0, 50) == 50) {
                 $this->prize = Prizes::$EXP_HAT;
             }
-            if (rand(0, 45) == 45) {
+            if (rand(0, 50) == 50) {
                 $this->prize = Prizes::$SANTA_HAT;
             }
             if (rand(0, 50) == 50) {
                 $this->prize = Prizes::$PARTY_HAT;
             }
-            if (rand(0, 40) == 40 && HappyHour::isActive()) {
+            if (rand(0, 50) == 50 && HappyHour::isActive()) {
                 $this->prize = Prizes::$JUMP_START_HAT;
             }
         }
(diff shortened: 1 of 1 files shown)


bls1999:
Can you stop with the malicious pulls?  This would create a botting exploit where botters could get the maximum amount of EXP before the race starts.  You're not fooling anyone.  DO NOT MERGE @jacob-grahn.

MoonWithSunshine:
@bls1999 No, there is no exploit that you speak of. The race still has to begin before one can finish race, and it takes your time from after you begin the race. It is bad to subtract by null values, which is why I added the check in.

bls1999:
There's no exploit right now, but merging this request would create one.  Honestly, it's so easy to pick out these malicious pulls.

multiplayer_server/rooms/Game.php
```diff
@@ -428,7 +428,10 @@ public function finish_race($player)
     {
         global $pdo;
 
-        if ($player->finished_race === false && !isset($player->race_stats->finish_time) && $player->race_stats->drawing === false && $this->begun === true) {
+        if ($player->finished_race === false && !isset($player->race_stats->finish_time)) {
+            if($this->start_time == null) {
+                $this->start_time = 0;
```
aaaaaa123456789: This is a bad default value for `start_time`, as it would cause inordinately large values for `$finish_time` in the line below. `$this -> start_time = microtime(true);` is a better way of handling this erroneous solution — after all, `start_time` shouldn't be null in the first place, so this error handler should rarely be triggered.

multiplayer_server/rooms/Game.php
```diff
@@ -428,7 +428,10 @@ public function finish_race($player)
     {
         global $pdo;
 
-        if ($player->finished_race === false && !isset($player->race_stats->finish_time) && $player->race_stats->drawing === false && $this->begun === true) {
+        if ($player->finished_race === false && !isset($player->race_stats->finish_time)) {
+            if($this->start_time == null) {
```
aaaaaa123456789: `if(!$this -> start_time)` is shorter.

bls1999:
@MoonWithSunshine I can't wait to hear your explanation for removing `$player->race_stats->drawing === false && $this->begun === true` :joy: :joy:

(If anyone is unaware, this would open the door for botters to be able to win the race before it starts, thereby gaining the maximum EXP possible.  The null check only solidifies this, as the level completed multiplier relies on `finish_time - start_time` to make a calculation. Setting `start_time` to 0 would give an exorbitant amount of EXP.)

Relokos:
What is the difference between botting and macroing anyway?

aaaaaa123456789:
Botting uses a custom-crafted PR2 client to interact with the server; macroing uses software to simulate user inputs on the official client.

In the context of this pull, it would allow a bot to rapidly send race start followed by race finish commands, thus finishing a race nearly instantaneously while still getting exp for the full 2 minutes. While it would br possible to exploit this from a macro as well, a macro would require a few seconds per race (because of various delays: the time it takes to load the level, start the race, hit the finish block, close the exp window...); a bot could just spam the commands 100 times per second.

Relokos:
Oh I see, thank you. @aaaaaa123456789

zazz2:
you must think @bls1999 is clueless or something because he's seen through every single attempt of yours to resurrect OP botting.

that being said, this person most likely has bots that take advantage of patched exploits like this (old or new). they're attempting to unpatch it in order to have a VERY unfair advantage over everyone else. remember to check everyone's pull requests CAREFULLY in order to ensure that a dangerous pull request like this is not merged. in response to @aaaaaa123456789, i don't think this can be exploited through a macro, but maybe i'm not seeing something.

one more thing @MoonWithSunshine, when the leaderboard.php pull request gets merged, we'll all get to see the secret accounts that you and your goons made and ranked up when these exploits weren't patched.

@jacob-grahn there is no issue with the code right now. do not merge this request.

aaaaaa123456789:
@zazz2 this pull would make _every_ race have a large finish time, so a macro could also exploit it by avoiding having to wait 2 minutes between races. It's just a lot slower than a bot, naturally; they'd still need some small delays for synchronization purposes, meaning that they'd be finishing a race every few seconds (instead of every 2 minutes and change).

bls1999:
@MoonWithSunshine Funny how you can push a commit but can't explain why you removed `$player->race_stats->drawing === false && $this->begun === true` :joy: :joy:

PR #346, opened 2018-04-01
Anti-botting safety checks

aaaaaa123456789:
Partly because I'm annoyed by the malicious pulls. These checks would prevent the tricks used by those bots from being exploited, even if they succeed.

### DIFF
--- a/multiplayer_server/rooms/Game.php
+++ b/multiplayer_server/rooms/Game.php
@@ -430,6 +430,7 @@ class Game extends Room
 
         if ($player->finished_race === false && !isset($player->race_stats->finish_time) && $player->race_stats->drawing === false && $this->begun === true) {
             $finish_time = microtime(true) - $this->start_time;
+            if ($finish_time > 31536000) $finish_time = 0; // if the race time is >1 year, set it to 0. This is just a glitch protection.
             $this->set_finish_time($player, $finish_time);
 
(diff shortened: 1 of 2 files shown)


multiplayer_server/rooms/Game.php
```diff
@@ -430,6 +430,7 @@ public function finish_race($player)
 
         if ($player->finished_race === false && !isset($player->race_stats->finish_time) && $player->race_stats->drawing === false && $this->begun === true) {
             $finish_time = microtime(true) - $this->start_time;
+            if ($finish_time > 31536000) $finish_time = 0; // if the race time is >1 year, set it to 0. This is just a glitch protection.
```
bls1999: You could probably even make this a day. I don't think there's ever been a level that takes a day to complete.

MoonWithSunshine: Nope. There are levels that take days and even months to complete.

aaaaaa123456789: The current UNIX timestamp is 1522 million and counting; 31.5 million is good enough to catch those botters. I set it stupidly large on purpose.

aaaaaa123456789: @MoonWithSunshine so a year is OK? Good.

jacob-grahn:
:+1:

PR #349, opened 2018-04-02
Remove final references to Lux, repurpose Moon Hat

bls1999:
Some small updates:
 - Remove all references to Lux, as it's no longer used.
 - Make the moon hat give double GP.
 - Add the time in which a level was defeated to the chat message during tournament mode.
 - Move some code in Player.php to let guests/users less than rank 3 use chat commands.

### DIFF
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -95,7 +95,6 @@ class Player
 
     public $status = '';
 
-    public $lux = 0;
     public $lives = 3;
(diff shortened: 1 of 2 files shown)


jacob-grahn:
> Add the time in which a level was defeated to the chat message during tournament mode

That's really slick!

jacob-grahn:
This is now running on Fred and Isabel if anyone is up for some testing.

PR #366, opened 2018-04-05
Enable referer check

Myst1cZ:
Enable referer check to prevent CSRF exploit that would allow to force change campaign.


jacob-grahn:
:+1:

PR #373, opened 2018-04-08
Prohibit backtick in race chats

bls1999:
This prevents socket commands other than `` chat` `` from being executed via the chat box. To take effect, this pull requires restarting all servers.

Thanks to @BrokenBulb78 for bringing this to my attention! \o/


BrokenBulb78:
Well done m8 ^^

Issue #370, opened 2018-04-07
One more column


bls1999:
@jacob-grahn Can you add `awarder_ip` to `contest_winners`? `varchar(100)` should suffice. This is it, I promise!

Relokos:
bls can i talk to you regarding an issue, ofc nothing to do with me. do you have an email where i can contact you

bls1999:
My email is bls1999@jiggmin2.com. Please keep issues on github on-topic.

jacob-grahn:
Done!

bls1999:
Thank you!  Hopefully everything works as intended. 🙏🏻

PR #374, opened 2018-04-08
Fix Contests Hub (last pull about this???)

bls1999:
I was having trouble trying to figure out how I'd even start to go about debugging these scripts.  Then I remembered that the DB queries are public, so I created a few tables in a database on my localhost in order to mimic the PR2 server DB's structure for proper testing.  I have good ideas sometimes.

Anyway, this all worked on my localhost so it should theoretically work on PR2's server.  Please add a column to `contest_winners` (as outlined in #370) before merging.

I'll fix those pesky IP searches next. :roll_eyes: :joy:


jacob-grahn:
Whew!

PR #377, opened 2018-04-08
Add HttpOnly to token cookie

Myst1cZ:
Make token cookie HttpOnly for extra security on the cookie.


Myst1cZ:
Closed: i need to get rid of that merge to my fork.

PR #382, opened 2018-04-09
Actually fix leaderboard, guild search

bls1999:
**Guild Search:**

This seemed to work on my localhost. It will hopefully have the same result here!

**Leaderboard:**

While debugging, I noticed that `users_select_top` was requiring `GROUP BY` to be present with the aggregate function `SUM()`.  When I grouped by `name`, it told me `pr2.rank` needed to be included since it was in the `SUM()` function.  It subsequently continued to tell me I needed more columns until I ended up with the monstrosity that is line 14.  From there, it seems to work okay in my test database, although I'm not sure if that's an effective method of fixing the problem. :thinking:

An alternative fix is to disable `ONLY_FULL_GROUP_BY` in the database to get rid of the `GROUP BY` line (14).  More info here: https://stackoverflow.com/questions/23921117/disable-only-full-group-by


jacob-grahn:
That is one hairy query, indeed

PR #383, opened 2018-04-09
Add /unkick for accidental kicks

bls1999:
Currently, the only way to get rid of an accidental kick is to restart the entire server. This pull enables trial/permanent mods and server owners (no temps) to remove accidental kicks using the /unkick command.

Edit: This pull also includes a fix for admins trying to demote offline users displaying "Error: Could not find a user with that ID."

### DIFF
--- a/multiplayer_server/LocalBans.php
+++ b/multiplayer_server/LocalBans.php
@@ -16,6 +16,21 @@ class LocalBans
     }
     
     
+    public static function remove($user_name)
+    {
+        $len = count(self::$arr);
+        for ($i=0; $i<$len; $i++) {
+            $ban = self::$arr[$i];
+            if ($ban->user_name == $user_name) {
+                array_splice(self::$arr, $i, 1);
+                return true;
+            }
+            continue;
+        }
+        return false;
+    }
+    
+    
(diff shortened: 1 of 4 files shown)


multiplayer_server/LocalBans.php
```diff
@@ -16,6 +16,21 @@ public static function add($user_name, $duration = 1800)
...
+            if ($ban->user_name == $user_name) {
+                array_splice(self::$arr, $i, 1);
+                return true;
+            }
+            continue;
+        }
+        return false;
+    }
+    
+
```
jacob-grahn: I figured there's got to be an easier way to remove a value from an array, and google revealed `array_diff`! We might be able to use a one liner here, but then we'd loose the true/false return
`self::$arr = array_diff(self::$arr, [$user_name])`

I don't really know why I set this up with user names instead of user ids. If a user logs in with different capitalization in their name, would they evade the kick?

bls1999: Upon testing, it seems that different capitalization does not influence a user's kick status.  It'll keep them out regardless of the capitalization of their name.  It would definitely be a good idea to convert it to usernames.

PR #384, opened 2018-04-09
Add htmlspecialchars to ban reason

Dex472:
Just so the ban message won't have html code.

### DIFF
--- a/http_server/fns/data_fns.php
+++ b/http_server/fns/data_fns.php
@@ -117,7 +117,15 @@ function format_duration($seconds)
 
 function get_ip()
 {
-    return $_SERVER['REMOTE_ADDR'];
+    $ip = "";
+    if (isset($_SERVER["HTTP_X_FORWARDED_FOR"])){
+        $ip = $_SERVER["HTTP_X_FORWARDED_FOR"];
+    }
+    else
+    {
+        $ip = $_SERVER['REMOTE_ADDR'];
+    }
+    return $ip
 }
 
 function check_value($value, $check_for, $yes = 'yes', $no = 'no')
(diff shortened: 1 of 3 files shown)


isokissa3:
X-Forwarded-For header should not be trusted outside private network, it can be easily spoofed

http_server/fns/data_fns.php
```diff
@@ -117,7 +117,15 @@ function format_duration($seconds)
 
 function get_ip()
 {
-    return $_SERVER['REMOTE_ADDR'];
+    $ip = "";
+    if (isset($_SERVER["HTTP_X_FORWARDED_FOR"])){
```
bls1999: This can be easily spoofed and would make our jobs exponentially harder.

http_server/fns/data_fns.php
```diff
@@ -117,7 +117,15 @@ function format_duration($seconds)
 
 function get_ip()
 {
-    return $_SERVER['REMOTE_ADDR'];
+    $useripaddress = "";
+    if (isset($_SERVER["HTTP_X_FORWARDED_FOR"])){
```
bls1999: This can be easily spoofed and would make our jobs exponentially harder.

Dex472:
Did the change to check for the ip make it any better?

bls1999:
Not really; the problem of `HTTP_X_FORWARDED_FOR` being easily spoofed is still present.

isokissa3:
🤦‍♂️

bls1999:
@Dex472 https://stackoverflow.com/questions/9221082/preventing-ip-spoofing

Dex472:
Ok, reversed the changes to get_ip()

jacob-grahn:
:+1:

PR #385, opened 2018-04-10
Add debugging for login.php

bls1999:
I'm pretty sure that the only way I can let myself see guests' login data without letting everyone else see it is to simply allow the first bit of my IP address to see it and no one else.  I imagine this will definitively let me figure out if there is something different sent to the server to prohibit guest accounts from earning Kong badges, and in turn, fix it (if possible).  I'll send you the value for `$BLS_IP_PREFIX`.

### DIFF
--- a/env.example.php
+++ b/env.example.php
@@ -42,4 +42,6 @@ $LOGIN_IV = 'there';
 
 $GUEST_PASS = 'this is the universal password for all guest accounts';
 
+$BLS_IP_PREFIX = 'no ddos plz';
+
 ?>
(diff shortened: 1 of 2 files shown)


Dex472:
Wouldn't this allow you to see other players login data aswell?

bls1999:
Only if I'm the one logging into them.  I wouldn't be able to see your login data, for example.

bls1999:
@Dex472 Part of the reason why I hate debugging the multiplayer server is because I'm not trying to see GET/POST data, so normal web debuggers like that one won't work.  I'm trying to see what's sent to the PR2 server after login.php authenticates the user.  Even if I was trying to see what was sent to login.php, it would be encrypted.

PR #390, opened 2018-04-10
One last change in guild_search.php

BrokenBulb78:
Changes I made:

- Always output the header so we don't need to call that function many times.
- Not throwing "No guild name or id." when someone is loading the page.
- Removed a center tag, there were 2 lol


BrokenBulb78:
sorryyyy forgot something lol

PR #392, opened 2018-04-11
Some more changes

BrokenBulb78:
- Added missing footer on backups page.
- Made links on main page open in new tab.

### DIFF
--- a/http_server/fns/output_fns.php
+++ b/http_server/fns/output_fns.php
@@ -46,10 +46,10 @@ function output_footer()
 
     <div id="footer">
         <ul class="footer_links">
-            <li><a href="//pr2hub.com/backups">Backups</a></li>
-            <li><a href="https://jiggmin2.com/forums/showthread.php?tid=19">Folding@Home</a></li>
-            <li><a href="//pr2hub.com/terms_of_use.php">Terms of Use</a></li>
-            <li><a href="https://jiggmin2.com/forums/showthread.php?tid=385">Rules</a></li>
+            <li><a href="//pr2hub.com/backups" target="_blank">Backups</a></li>
+            <li><a href="https://jiggmin2.com/forums/showthread.php?tid=19" target="_blank">Folding@Home</a></li>
+            <li><a href="//pr2hub.com/terms_of_use.php" target="_blank">Terms of Use</a></li>
+            <li><a href="https://jiggmin2.com/forums/showthread.php?tid=385" target="_blank">Rules</a></li>
         </ul>
     </div>
 </div>
--- a/http_server/fns/output_fns.php
+++ b/http_server/fns/output_fns.php
@@ -64,22 +64,22 @@ function output_mod_navigation($formatting_for_admins = true)
 
     <p>
         <b>
-            <a href="//pr2hub.com/mod/reported_messages.php">Reported Messages</a>
+            <a href="//pr2hub.com/mod/reported_messages.php" target="_blank">Reported Messages</a>
             -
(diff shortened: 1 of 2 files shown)


http_server/www/backups/backups.php
```diff
@@ -107,6 +107,8 @@
     foreach ($backups as $row) {
         echo "<p>$row->date: <b>".htmlspecialchars($row->title)."</b> v$row->version <a href='?action=restore&backup_id=$row->backup_id'>restore</a></p>";
     }
+    
+    output_footer();
```
bls1999: Maybe put this in a finally block after the try/catch? so like this:

```
try {
    // stuff
} catch (Exception $e) {
    // more stuff
} finally {
    output_footer();
    die();
}
```

http_server/fns/output_fns.php
```diff
@@ -46,10 +46,10 @@ function output_footer()
 
     <div id="footer">
         <ul class="footer_links">
-            <li><a href="//pr2hub.com/backups">Backups</a></li>
-            <li><a href="https://jiggmin2.com/forums/showthread.php?tid=19">Folding@Home</a></li>
-            <li><a href="//pr2hub.com/terms_of_use.php">Terms of Use</a></li>
-            <li><a href="https://jiggmin2.com/forums/showthread.php?tid=385">Rules</a></li>
+            <li><a href="//pr2hub.com/backups" target="_blank">Backups</a></li>
```
bls1999: I don't think this is needed.

http_server/fns/output_fns.php
```diff
@@ -64,22 +64,22 @@ function output_mod_navigation($formatting_for_admins = true)
 
     <p>
         <b>
-            <a href="//pr2hub.com/mod/reported_messages.php">Reported Messages</a>
+            <a href="//pr2hub.com/mod/reported_messages.php" target="_blank">Reported Messages</a>
```
bls1999: I'm not sure what else to say other than "please no"

http_server/fns/output_fns.php
```diff
@@ -64,22 +64,22 @@ function output_mod_navigation($formatting_for_admins = true)
...
+            <a href="//pr2hub.com/admin/player_deep_info.php" target="_blank">Update Account</a>
             -
-            <a href="//pr2hub.com/admin/guild_deep_info.php">Update Guild</a>
+            <a href="//pr2hub.com/admin/guild_deep_info.php" target="_blank">Update Guild</a>
             -
-            <a href="//pr2hub.com/admin/set_campaign.php">Set Custom Campaign</a>
+            <a href="//pr2hub.com/admin/set_campaign.php" target="_blank">Set Custom Campaign</a>
             -
-            <a href="//pr2hub.com/admin/admin_log.php">Admin Action Log</a>
+            <a href="//pr2hub.com/admin/admin_log.php" target="_blank">Admin Action Log</a>
```
bls1999: In all seriousness, adding `target="_blank"` would be really annoying for us.  I use CTRL + click when I need to open a link in a new tab.  It's not really good for navigation links.

LP-PR2:
die();

jacob-grahn:
Usually people don't expect links to open in new tabs. The argument could be made that the links on pr2hub should open new tabs as a way to stop people from accidentally loosing unsaved work in the level editor, but really I'd rather just stick with normal links.

BrokenBulb78:
Alright, I removed `target="_blank"` from the mod/admin links and I put the footer in a finally block after the try/catch. @jacob-grahn @bls1999

PR #394, opened 2018-04-11
Fix referer check bypass exploit

Dex472:
Fix for exploit to bypass referer check (can be used to run member/moderator/admin actions without confirmation depending on targeted user privileges.)


jacob-grahn:
Good catch!

PR #399, opened 2018-04-12
Fix for TokenEater force logout exploit

Dex472:
Fix for the exploit https://github.com/Dex472/TokenEater

### DIFF
--- a/http_server/www/logout.php
+++ b/http_server/www/logout.php
@@ -29,7 +29,7 @@ try {
     rate_limit('logout-'.$ip, 60, 10, 'Only 10 logout requests per minute per IP are accepted.');
 
     if (is_from_game() !== true){
-     throw new Exception("For security reasons logout can only be performed from game");   
+     throw new Exception("It looks like you're not using PR2 to log out. For security reasons, you may only log out from a PR2 client.");   
     }
     
     if (isset($_COOKIE['token'])) {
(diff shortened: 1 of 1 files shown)


http_server/www/logout.php
```diff
@@ -5,13 +5,33 @@
 
 header("Content-type: text/plain");
 
+function is_from_game(){
```
bls1999: The formatting is off.

http_server/www/logout.php
```diff
@@ -5,13 +5,33 @@
...
+  {
+      return false;   
+  }
+ }
+ else
+ {
+    return false;
+ }
+}
+
```
bls1999: The thought of using `HTTP_X_REQUESTED_WITH` crossed my mind; However, this too can be easily spoofed.

http_server/www/logout.php
```diff
@@ -5,13 +5,33 @@
...
 $ip = get_ip();
 
 try {
     // rate limiting
     rate_limit('logout-'.$ip, 5, 2, 'Please wait at least 5 seconds before attempting to log out again.');
     rate_limit('logout-'.$ip, 60, 10, 'Only 10 logout requests per minute per IP are accepted.');
 
+    if (is_from_game() !== true){
+     throw new Exception("For security reasons logout can only be performed from game");
```
bls1999: Change language to keep consistent with other third-party site warnings. i.e. "It looks like you're not using PR2 to log out. For security reasons, you may only log out from a PR2 client."

jacob-grahn:
Does flash always include the HTTP_X_REQUESTED_WITH header with outbound requests?

Dex472:
@jacob-grahn Just checked it and it does send it everytime the request's origin is from the pr2 client.

jacob-grahn:
:+1:

PR #400, opened 2018-04-12
Simplify is_from_game()

bls1999:
There's really no way to avoid this monstrosity of a line while keeping PSR-2 compliance without breaking it into a few lines.  Additionally, doing it all in one if statement is more efficient when both versions of the code would put the lines over 80 characters per line anyway.

Also, removed some unnecessary if/else blocks and consolidated our return boolean into one variable.


bls1999:
Wrong repo!

PR #401, opened 2018-04-13
Add updated_time to artifact_hint.txt output

bls1999:
Just an additional bit of information to be output to artifact_hint.php


jacob-grahn:
:+1:

PR #403, opened 2018-04-14
Slight fix for /mod/player_info.php

bls1999:
Prevent the script from blowing up if the player doesn't have an entry in the PR2 table.


jacob-grahn:
:+1:

PR #404, opened 2018-04-14
Fix tournament win times

bls1999:
Fixes the issue where the third decimal digit always displays as zero on scores being broadcasted to main with tournament mode on.

This should be fairly easy to move to lint, once ready.


jacob-grahn:
:+1:

PR #406, opened 2018-04-15
Check if guild exists before attempting to create

bls1999:
Right now, if you attempt to create a guild and it exists, you can't try again for another hour due to how the rate limiting is set up.  This pull checks if a guild name exists before adding to the rate limit and attempting to create a new guild, thereby fixing the issue.


jacob-grahn:
Looks good, thanks!

PR #410, opened 2018-04-17
Referrer/from game check changes

bls1999:
- Update the messages to reflect the action that is attempting to be performed by the client
 - Add referrer checking to `message_delete.php` and `submit_rating.php`
 - Simply say "incorrect referrer" on mod action failing due to referrer check


jacob-grahn:
Cool, thanks for cleaning that up.

PR #411, opened 2018-04-17
Fixes to Guild Search

bls1999:
- Members now display their individual GP stats rather than the guild's GP stats
 - Center everything


jacob-grahn:
:+1:

PR #413, opened 2018-04-18
Revert is_from_game fix in logout.php

bls1999:
It's throwing an error every time someone logs out.


Dex472:
@bls1999 What error it throws?

bls1999:
@Dex472 Error: It looks like you're not using PR2 to log out. For security reasons, you may only log out from a PR2 client.

Dex472:
I think i figured out why this happens, will make pull request as soon i get to fixing it.

bls1999:
@Dex472 could you explain?

jacob-grahn:
I'll revert this for now so people can log out until we've got a fix.

PR #417, opened 2018-04-18
Change "!==" to "==="

Dex472:
Changed "!==" on strpos to "===" to correctly check is "ShockwaveFlash/" in the X-Requested-With header because "!==" was checking if it's not in string so this should fix the error when using this on logout.
Referer check wasn't needed there because it doesn't really have an use on this case.

### DIFF
--- a/http_server/fns/data_fns.php
+++ b/http_server/fns/data_fns.php
@@ -4,9 +4,10 @@
 function is_from_game()
 {
     $req_with = default_server("HTTP_X_REQUESTED_WITH");
+    $referer = default_server("HTTP_REFERER");
     $is_from_game = false;
 
-    if (!empty($req_with) && strpos($req_with, "ShockwaveFlash/") === 0) {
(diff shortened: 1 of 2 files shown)


bls1999:
The point of checking if the referrer is empty is so people can go to the pr2hub.com/logout.php link directly from their browser.  We discussed this before your original pull was merged.  I have used this link many times to delete my tokens.

Dex472:
@bls1999 Oh, sorries, i forgot about that.

Dex472:
It's really confusing that check_ref has been renamed to require_trusted_ref or maybe it is is_trusted_ref but i have no idea how that's even supposed to work since the is_trusted_ref returns the strpos checks instead of bool value so i won't add that since there's no real way to know it would even work after that.

bls1999:
Oh, you're right.  The best solution would probably be to separate `require_trusted_ref` and the actual referrer check.

jacob-grahn:
Thanks guys!

PR #418, opened 2018-04-18
Fix errors

bls1999:
- Removed some dead require /mod/mod_fns.php links 
 - Fixed a syntax error in /contests/award_prize.php
 - Added a missing require link in /admin/contests/edit_contest.php
 - Moved problematic HTML tag in guild_search.php
 - Moved the referrer check in login.php to the beginning of the file, rather than after the user has logged in.


jacob-grahn:
:+1:

PR #420, opened 2018-04-19
FIX FOLDING AT HOME

bls1999:
UPDATE:

So, funny story.  I spent 7.5 hours tonight (midnight to 7:30am) reworking the code to `fah_award_prizes_fns.php` trying to minimize the likelihood of failure.  It turns out, the folding issue was due to spaces not replacing underscores correctly.  I went ahead and corrected that in `fah_fetch_stats.php`.  You may want to just truncate the folding database's `stats` table and then run the scripts to fetch data and award prizes manually, since there will be tons of entries in the `stats` table with underscores instead of spaces.

OLD DESCRIPTION:

Since tons of users have messaged Eternal and me about their folding prizes not showing up, I've made a script that should let us trigger an update of those stats for one specific user.

I did some testing on my localhost, so theoretically it should work; However, from past experiences with this repo, it probably won't work on the first try.


jacob-grahn:
:+1:

PR #421, opened 2018-04-19
Allow admins to restart all servers at once

bls1999:
Just in case.  We can already restart individual servers, so if we need to restart all of them, we can just use this script once per hour instead of typing the command in every server one-by-one.  Extra security by passing the token via POST instead of reading it in the cookie.


jacob-grahn:
Cool :sunglasses:

PR #424, opened 2018-04-19
Add page leaving confirmation in index page

BrokenBulb78:
I closed the other pull because it was completely messed up, this one will fix the issue. The original commit was created by @bls1999.

### DIFF
--- a/http_server/www/index.php
+++ b/http_server/www/index.php
@@ -1,16 +1,23 @@
 <?php
+
 require_once __DIR__ . '/../fns/output_fns.php';
+
 output_header('Platform Racing 2');
-?>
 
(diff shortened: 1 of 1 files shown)


jacob-grahn:
I'm just not a fan of opening a confirmation popup. I can't remember the last time a site did that in a useful way, so I associate confirmation popups with scammy websites.

bls1999:
@jacob-grahn What if there was the option to disable the popup?  I'm thinking if you want it in the first place, you can check a little box directly below the client.  As long as that box is checked, you'll get warned before leaving.

bls1999:
@BrokenBulb78 :man_shrugging: sorry bud

PR #434, opened 2018-04-22
Fix users registering and streamline registration process

bls1999:
- Move all of the individual steps to functions called in `/fns/register_user_fns.php`.
 - Add check for incomplete user data (missing `pr2` row) and if found, insert a row on login.


jacob-grahn:
That is one clean PR :+1:

bls1999:
Lol thanks

PR #436, opened 2018-04-23
Changes to text emotes

bls1999:
- Add `:thinking:` and `:lol:` (as well as aliases for the two)
 - Make the emotes case-insensitive (i.e. replace the text regardless of `:shrug:` or `:SHRUG:`)
 - Add `/emotes`, which explains the syntax for the different emoticons.

Unicode has evolved, and as a result, emojis are available on more devices than ever before.  I googled how to put unicode characters into a php file, and it just said to paste them straight in. 🤷‍♂️

### DIFF
--- a/multiplayer_server/Player.php
+++ b/multiplayer_server/Player.php
@@ -279,8 +279,10 @@ class Player
 
         //special text emotes
         if ($room_type == 'c') {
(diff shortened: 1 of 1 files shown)


jacob-grahn:
:+1:

PR #437, opened 2018-04-23
Some fixes

BrokenBulb78:
**hint.php** - Added `require_once` for the `data_fns.php`. This missing line made the whole script break.

**backups.php** - Added the missing footer.

**pr2hub.css** - removed `.above_game_ad` CSS. It's useless since Jacob removed the ad from the page.


jacob-grahn:
:+1:

PR #438, opened 2018-04-24
Convert ensure_awards to fn, add to hourly cron

bls1999:
I figured since the script needed fixing anyway, it would make sense just to add it to the hourly cron while we're at it. 🤷‍♂️


jacob-grahn:
Looks good!

PR #443, opened 2018-04-26
Guest checks for assorted files in the http_server

bls1999:
When I end up submitting another pull to test a new fix for Kong badges, these things will need to be disabled so that guests don't have access to all the member-only stuff.


jacob-grahn:
:+1:

PR #444, opened 2018-04-27
Complete CRONpocalypse

bls1999:
Things accomplished:
 - Converted the stray cron jobs (outside minute, hourly, daily, and weekly) to functions and added them to one of the four
 - Moved some things from daily to weekly
 - Converted `echo_line()` to `output()` and moved `output()` to `data_fns.php`
 - Made FAH completely independent of the FAH database
 - Modified `queries/misc/all_optimize.php` to get a fresh list of tables to optimize each time it's run

I would've changed `output` to `echo_line`, but it looked like `echo_line` was only used in one place, whereas `output` was used in a tooooon of files.


jacob-grahn:
It has been a fine CRONpocalypse

bls1999:
@jacob-grahn indeed sir!

PR #445, opened 2018-04-27
Remove code related to fah db

bls1999:
Weeeeeee


jacob-grahn:
Tryin' to remove my cars, sheesh.

PR #449, opened 2018-04-29
Housekeeping

bls1999:
A bunch of small updates that fix a bunch of small things.  I also improved some of the outputs from the socket server.  Take a look and see if things look good; I tested it all locally, so it should work.


jacob-grahn:
Looks good to me :+1:

PR #450, opened 2018-04-29
Remove the old sitelock

BrokenBulb78:
Since we can use the `require_trusted_ref` function, the old sitelock is useless, which can be bypassed anyway.


BrokenBulb78:
Yup, it's different. `$origination_domain` checks if the client domain variable is `local`. You can change that variable to whatever you want to bypass the sitelock. So that's why I made this pull request, this code's useless now.
image

jacob-grahn:
I'm cool with taking this site lock out. People clearly know how to get around it anyway. :+1:

PR #453, opened 2018-04-30
exp sanity check

famelegend1:
doing this very quickly. this is another botting sanity check and exploit sanity check. eventually i may want to convert this to a rate limiting mechanism, but i have to work that out.

### DIFF
--- a/multiplayer_server/rooms/Game.php
+++ b/multiplayer_server/rooms/Game.php
@@ -626,8 +626,9 @@ class Game extends Room
                     $player->write('award`Artifact Found!`+ ' . number_format($artifact_bonus));
                 }
             }
-            if($tot_exp_gain > 100000)
-                $tot_exp_gain = 0;
+            if($tot_exp_gain > 100000) {
+                $tot_exp_gain = 0; //resets unrealistic exp
+            }
             
             //---
             $player->incExp($tot_exp_gain);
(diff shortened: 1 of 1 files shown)


famelegend1:
oh no did i mess up?

famelegend1:
i can't wait to hear what the rodents have to say about my adding useful code pull request. @bls1999

bls1999:
"Update Game.php"

famelegend1:
what happens when i don't add a title. malicious pull request confirmed.

bls1999:
lmao
In all seriousness, this is a good pull. @jacob-grahn

RealShadowZ:
What's the point of this. You can't even get that much Exp in a race so it's just useless

bls1999:
There's no point in impersonating other users to attempt to steal their accounts, but you do that anyway! You're in no place to comment on the validity of something on the grounds of there being no point in doing it. @RealShadowZ

aaaaaa123456789:
@RealShadowZ that's the point! You can't get that much exp in a race, so if you do, you must be abusing an exploit.

famelegend1:
any response from you, mark? @RealShadowZ

bls1999:
@jacob-grahn #456 combines this and #455, as well as adding a few other places to check.

PR #455, opened 2018-04-30
Rank Sanity check

famelegend1:
eventually i want to do a check like this in login.php, but im a little uneasy about touching that right now. this should for now patch an exploit that szr used

### DIFF
--- a/multiplayer_server/rooms/Game.php
+++ b/multiplayer_server/rooms/Game.php
@@ -535,7 +535,7 @@ class Game extends Room
             for ($i = $place+1; $i < count($this->finish_array); $i++) {
                 $race_stats = $this->finish_array[$i];
                 $exp_gain = ($race_stats->rank+5) * $time_mod;
-                if ($race_stats->rank+5 > 135) {
+                if (($race_stats->rank+5) > 135) {
                     $exp_gain = 0;
                 }
                 $exp_gain = ceil($this->appyExpCurve($player, $exp_gain));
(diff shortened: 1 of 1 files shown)


famelegend1:
yeah i sort of have my doubts on this, but can fix it if something goes wrong

bls1999:
My worry is that it will crash the server when encountered.  It would be better to test this before it goes in.

famelegend1:
@bls1999 results from your localhost testing?

famelegend1:
@Dex472 suggested that i use intval here. agree?

bls1999:
I took a nap lol, I'll test it now

aaaaaa123456789:
I'm not all too sure of the precedence between `+` and `<`; it would be better to add parentheses (`if (($race_stats -> rank + 5) > 135)`.

Other than that, I don't see why it should cause any problems.

famelegend1:
im not sure on the hierarchy but yeah that might be an interesting idea.

RealShadowZ:
Don't merge this because it will crash the server based on where it's getting the values from. Dangerous pull request

bls1999:
Don't worry @RealShadowZ, I have a pull coming that won't let anyone log in or join any races with an active rank more than 100.

famelegend1:
@RealShadowZ screaming "dangerous pull request"... like you and the other rodents haven't been making dangerous pull requests. i could provide a list of them if you want?

bls1999:
@jacob-grahn #456 combines this and #453, as well as adding a few other places to check.

PR #456, opened 2018-05-01
Add various rank/exp sanity checks

bls1999:
This pull includes code that @famelegend1 wrote in #453 and #455, slightly modified.  A huge thank-you to him for inspiring the changes in ab19b37. :pray:

Socket Server:
 - Check for rank 100 or above (except for Fred) on login in `Player.php`
 - Check for rank 100 or above (except for Fred) when entering a race, d/c if above rank 100
 - Check for a rank 100 or above when awarding opponent bonus EXP, and set to 0 if found
 - Fix php warnings being sent to the console when `` loginFailure` `` is sent from `Player.php` (i.e. server full, rank 100 or above check on login)
 - Fix a small bug that would crash the server if nothing was set for the hat, head, body, or feet values

HTTP Server:
 - Check for rank 100 or above on login in `login.php`

README.md:
 - Added @famelegend1

### DIFF
--- a/README.md
+++ b/README.md
@@ -20,6 +20,7 @@ To submit new ideas, post a thread in the [PR2 Suggestions](https://jiggmin2.com
 - BrokenBulb78
 - Dex472
 - EternalUF
+- famelegend1
 - fourst4r
 - Grovbi
 - isokissa3
--- a/http_server/www/login.php
+++ b/http_server/www/login.php
@@ -186,6 +186,12 @@ try {
     if ($rt_available < $rt_used) {
         $rt_used = $rt_available;
     }
+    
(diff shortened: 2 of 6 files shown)


famelegend1:
hey, thanks :)

@bls1999 you definitely know this repo more inside and out than i do. thanks for improving what i started.

@jacob-grahn this is worth merging. it prevents from using exploit r127 accounts to macro. leaderboard.php didn't catch all of them because some are apparently all in rank tokens.

i'll do the honors by giving the famous jiggmin thumbs up: :+1:

jacob-grahn:
Here's another :+1: for you

PR #471, opened 2018-05-06
Fix logins with more than 5 rank tokens

bls1999:
`login.php` currently doesn't let anyone with more than 5 used tokens log in. Also, Fred couldn't log in due to the rank checks. This fixes that.


jacob-grahn:
:+1: thanks

PR #472, opened 2018-05-07
Fix random song in levels

bls1999:
The typecast to an integer at the introduction of the `$song` variable messed up the saving of a random song, because the client interprets an empty `song=` value as a random song and `song=0` as "None".  This pull makes it so the song string (song ID; what's passed in `$_POST['song']`) is only typecast to an integer right before adding to the database.


jacob-grahn:
:+1:

PR #475, opened 2018-05-08
Add alternative right click menu file

Dex472:
Alternative right click menu with links to all pages that are also missing from bottom at page.


Dex472:
Invalid file name, don't merge.

PR #476, opened 2018-05-08
New right click menu

Dex472:
Alternative right click menu that can optionally be enabled on pr2hub for menu with missing page links.

Top notification showing how to enable:
https://gyazo.com/f1c360d809c5840469308293f68b7f36

Enabled menu when right clicking on page (appears to mouse position):
https://gyazo.com/313caf8c60189b207452b8097064f779

Preview of background change option:
https://gyazo.com/b10bac5c4386afeeffca947d4afe9b24

All of the things in menu is in javascript so it's only clientside change on background aswell so there's no risk of serverside code running. #

### DIFF
--- a/http_server/fns/output_fns.php
+++ b/http_server/fns/output_fns.php
@@ -129,6 +129,7 @@ function output_header($title = '', $formatting_for_mods = false, $formatting_fo
     <title>PR2 Hub - <?php echo $title; ?></title>
     <link href="//pr2hub.com/style/gwibble.css" rel="stylesheet" type="text/css" />
     <link href="//pr2hub.com/style/pr2hub.css" rel="stylesheet" type="text/css"/>
+    <script src="//pr2hub.com/pr2hubmenu.js" type="text/javascript"></script>
     <?php if ($formatting_for_mods) { ?>
         <script src="https://code.jquery.com/jquery-latest.min.js"></script>
         <script src="https://malsup.github.io/jquery.form.js"></script>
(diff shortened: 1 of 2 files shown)


BrokenBulb78:
I have some changes for it, check your repo.

bls1999:
As I said in DMs, I will not be approving this pull.  I use the stock right-click menu all the time.  Additionally, having a banner over the top on every page is rather sloppy and unneeded.

Don't get me wrong, I think this is a great pull with good features; However, I believe the right-click menu is just not the way to go.  I suggested making it a hover or click-to-reveal menu in the footer and was ignored.  I stand by that suggestion.  Obviously @jacob-grahn will make the final decision, but I can't approve this pull request.

Dex472:
@bls1999 This context menu starts disabled when going to the page and can be easily switched on and off

bls1999:
@Dex472 It's not about being able to be easily switched on and off; As I said privately, anything that modifies base browser functions is sketchy. The message on the top of every page is intrusive.

I asked privately why you feel it would be better in a context menu as opposed to a hover/click-to-reveal menu, and your answers were:
 - It wouldn't need to be added to every page
 - Only right-click returns the correct position of the cursor (for the pop-up at mouse position)
 - The right-click menu has more of a wow-factor

My responses:
 - Adding a top-of-page navigation or footer link, and calling the js file within `<head></head>` accomplishes the same thing.
 - Adding a hover or click-to-reveal event accomplishes the same thing.
 - While it may be "fancier" than a hover/click-to-reveal menu, fancy is not always better.

I would much rather have a click-to-reveal (or plain, no javascript) navigation menu at the top of the page, or a button in the footer that reveals a menu onmouseover or onclick.  Either of these options would be significantly less code and would not require a sloppy message at the top telling you how to enable something that disables browser functionality.

Dex472:
@bls1999 I guess i'll do the top navigation menu button later today then.

PR #478, opened 2018-05-09
Fixes to player_info.php and account_change_email.php

bls1999:
Some things broke in player_info.php after adding the GET parameter `name` to `/mod/player_info.php`.  This update fixes those broken things.

Also, a variable was misnamed in `account_change_email.php`. That has been fixed as well.


jacob-grahn:
:+1:

PR #480, opened 2018-05-09
Add top left corner menu

Altex1:
New menu for pr2hub with the missing page links.

Preview of the background change feature in it:
https://gyazo.com/c2ad9c34db7adf901edb3cde93762a8d


bls1999:
Good pull!  I have some changes I'm going to submit to your repo, but overall this is really really great.

One question: is the menu button stationary at the top of the page, or does it move with the scrolling?

jacob-grahn:
The background changer is cool, but I wonder if it introduces any security problems.

Altex1:
@jacob-grahn It doesn't cause any security risks because the change happens clientside, even if someone entered malicious image to it, then it would only run on their browser and no one else would get affected.

jacob-grahn:
Hey @Altex1, this is really cool! I think adding javascript to this simple page is a bit of overkill though.

PR #484, opened 2018-05-14
Organipocalypse pt. 3

bls1999:
Alright... this is the big one.

First thing's first... **_This pull REQUIRES a server configuration edit._**
Without this change, EVERY FILE in platform-racing-2-server will break.

The goal is to tell nginx to load config.php every time a php command or client request is made.
Add the following line within the "location ~ \.php$" block of your server configuration:

```
fastcgi_param PHP_VALUE "auto_prepend_file=/full/path/to/this/file/config.php";
```

Replace "/full/path/to/this/file" with the full path to config.php (in the root of the repo).

For more information, see these resources:
https://stackoverflow.com/questions/26192274/replacement-for-php-htaccess-values-in-nginx
https://stackoverflow.com/questions/14884439/automatically-load-a-config-php-file-for-all-pages-before-anything-else

In doing this, we can load scripts without requiring them right off the bat. We now also have some useful user-defined constants that we can use instead of using `__DIR__` all the time. A super special added bonus we get by doing this is easier path/organization changes (i.e. we want to move `queries` back to `http_server`; we simply move the folder and change the path's constant).

___

*_Changes!_*

Common:
 - Convert all `require_once` calls to use the constants
 - CRON directory moved here
 - Queries directory moved here
 - s3_connect.php moved here
 - Socket management moved into folder within common called /manage_socket/

Socket Server:
 - Condense all process functions into two files:
    - process_fns.php
    - vault_fns.php
 - Client functions; either:
    - Move into `/fns/client/`, or
    - Condense into `/fns/client/client_misc_fns.php`
 - Add new process functions (`process_message_player`, `process_activate_happy_hour`)
 - Add new socket management functions (`restart_server`, `restart_servers`, `check_servers`)
 - Rework existing socket management functions (improve outputs, consolidate multiple socket-related cron jobs to functions inside the normal cron folder)
 - Add /priors for lazy mods (looking at @a7x3 on this one). I tested and it seems to load the data almost immediately, so I don't think we need to worry about the server slowing down.

HTTP Server:
 - Completely rework/organize the fns directory
 - Move all files that contain functions from `/www/` to `/fns/`
 - Add `die()` at the end of all scripts to make sure the file is good and dead at the end of execution

To-Do:
 - Improve socket server responses when using `talk_to_server`, outputs, etc.
 - [Possibly] arrange the PR2 class files into folders (haven't decided on this one yet; do you have a preference?)
 - I had another good idea but of course I have forgotten it. Oh well; if it was a prolific idea, it should come back to me at some point. `¯\_(ツ)_/¯`

Personally, I see the organipocalypse as an ongoing effort to keep the repo organized. I will technically be finished with the organipocalypse at some point; However, it's up to us (the contributors) to keep the repo organized so we don't need another organipocalypse.

Finally, @jacob-grahn, If anything I've done makes you uneasy, let me know what I can do to change it. I don't want to twist your arm here; it's your code so you should decide how it's structured.


jacob-grahn:
Cool, I'll do some homework on this one. :smile:

bls1999:
That could definitely work! The only thing I was skeptical about was loading a bunch of unneeded functions on each request. Additionally, we'd have to make sure no two function names are the same throughout the entire repo to prevent errors.

bls1999:
@jacob-grahn ^^

jacob-grahn:
I went a little crazy and removed all but one of the die() calls. I also merged a lot of try / catch blocks together, which will result in `output_header` being called twice for some errors. The `output_*` functions need to get overhauled or replaced though, so I'm not too concerned about it in the short term.

PR #493, opened 2018-05-20
Organipocalypse pt. 3.5 (fix errors caused by pt. 3)

bls1999:
Some things broke. This fixes those things before continuing.

___

A couple of clarifications re: your vision (MrMaestro?) for how things will proceed:
 - What did you mean by "The `output_*` functions need to get overhauled or replaced"?
 - What's the status of the Templocalypse?

Also wanted to make sure you saw my response to this (posted by you on #484):
> What if we change all_fns.php to actually include all functions, and then set it up to get auto-loaded with each request? We could remove about a billion require_once statements from the repo.

My response:
> That could definitely work! The only thing I was skeptical about was loading a bunch of unneeded functions on each request. Additionally, we'd have to make sure no two function names are the same throughout the entire repo to prevent errors.
>
>If you're okay with all of this, then I'll get to it. To clarify: would you want all_fns.php to include all queries and both the multiplayer and http server functions, only both the multiplayer and http server functions (no queries), or just the contents of the http_server/fns/ folder? If we're going with options one or two, we could probably move both multiplayer_server/fns and http_server/fns into common and make http_server the public_html folder (instead of http_server/www/).

### DIFF
--- a/common/manage_socket/socket_manage_fns.php
+++ b/common/manage_socket/socket_manage_fns.php
@@ -258,13 +258,19 @@ function talk_to_server($address, $port, $salt, $process_function, $receive = tr
 function poll_servers($servers, $message, $output = true, $server_ids = array())
 {
     $results = array();
+    $query = array();
 
     foreach ($servers as $server) {
-        if (count($server_ids) == 0 || array_search($server->server_id, $server_ids) !== false) {
-            $result = talk_to_server($server->address, $server->port, $server->salt, $message, $output);
-            $server->command = $message;
-            $server->result = json_decode($result);
-            $results[] = $server;
+        $id = (int) $server->server_id;
+        $query = new stdClass();
+
+        if (count($server_ids) == 0 || array_search($id, $server_ids) !== false) {
+            $result = (string) talk_to_server($server->address, $server->port, $server->salt, $message, $output);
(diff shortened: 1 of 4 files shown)


jacob-grahn:
> I had two sets of try/catch blocks in lots of files that required a mod/admin token because otherwise, you would be met with this ugly sight. We can keep it like that, but it definitely looks neater to visitors who stumble upon pages to which they don't have access.
> What did you mean by "The output_* functions need to get overhauled or replaced"?

The goal is to run all of the logic first, and then render a page based on the result.  The system I naively set up when creating pr2 intermingles the logic and the page rendering. That complicates pretty much everything, and it's what lead to difficulties like needing multiple try / catch blocks just to show a decent error page.

This is more like what we want:

```php
$result = do_logic()
render_page($result)
```


>If you're okay with all of this, then I'll get to it. To clarify: would you want all_fns.php to include all queries and both the multiplayer and http server functions, only both the multiplayer and http server functions (no queries), or just the contents of the http_server/fns/ folder? If we're going with options one or two, we could probably move both multiplayer_server/fns and http_server/fns into common and make http_server the public_html folder (instead of http_server/www/).

Including `common` and `https_server/fns` should(?) take care of all of the scripts that the http server uses. Including the multiplayer functions shouldn't be needed.

As always, thanks for the pull! :+1:

PR #498, opened 2018-05-25
Fix for /hh status when triggered by random_hour

### LINKED ISSUE #474: /hh command does not work correctly.
image
This screenshot was took 19 minutes into the happy hour and it says there is -94214 seconds left on the happy hour which is -1570.2333 minutes. 
(Happy hours still do last an hour if anyone was confused)

Also seeing as now there is only 5 non guild servers instead of 9 that can have a happy hour on them shouldn't the chance of a happy hour be increased?

bls1999:
This should fix #474.


jacob-grahn:
:+1:

PR #502, opened 2018-06-02
Fix guest logins

bls1999:
I just reverted the guest login procedure to what it was before I attempted to fix the Kong badges.

Also, I added a line to `process_register_login()` in the socket server code that checks if the guest is online already. If so, it won't disconnect the guest that's online but it'll disconnect the person logging in. This way, people who know the guest password won't be able to troll others who are just visiting.

Edit: I also added a check for a direct guest account login in login.php.


jacob-grahn:
:+1: Looks good, sorry for my slowness

PR #505, opened 2018-06-25
Fix automatic artifact prize awards

bls1999:
The artifact function's set first finder function isn't awarding prizes properly because it's looking for a false or null value instead of 0 for the first finder. This pull fixes that and tidies some other potential problem points as well.


jacob-grahn:
:+1:

PR #507, opened 2018-06-29
Fix false positive

bls1999:
Sometimes on logout, users will get an error saying they can only log out from the PR2 client. This check was originally put in place to prevent a CSRF exploit, but seeing as it creates false positives and fixes an issue that is non-destructive in the first place, I think it should be retired.


jacob-grahn:
:+1:

PR #509, opened 2018-07-03
Limit publishing to newest

bls1999:
We've had an issue in the past with level spam. As a fix, [Stxtics suggested](https://jiggmin2.com/forums/showthread.php?tid=851) that there be a limit to the number of levels that a user can publish to page 1 of newest at any given time. This pull adds that limit while still letting the user save their level.

Side note: I noticed that when a level gets unpublished, the level remains in `pr2_new_levels`. To make this pull work, I needed to count the number of levels from that table. I've added a function to `upload_level.php` and `remove_level.php` which delete the level from `pr2_new_levels` if it exists in that table.


jacob-grahn:
👍

PR #510, opened 2018-07-07
Fix FAH pt. 79

bls1999:
Oh, you know. Folding at Home decided to update their fast teampage URL for the first time since PR2 came out, so it broke yet again.

I'd like to be able to go a year without this breaking, even if it's not my fault in the first place. That would be nice.


famelegend1:
LOL

bls1999:
Oof, wrong button.

jacob-grahn:
Nôöòóøōõ

Well, thanks for fixing it!

bls1999:
Sure thing @jacob-grahn!

PR #514, opened 2018-07-13
Fix server crash on part overflow

bls1999:
The old code I had here didn't do what I set out to accomplish it doing. Adding a `foreach` block to the existing `if` block seems to have done the trick.


bls1999:
@famelegend1 https://youtu.be/4Ya42tbu3ZA

jacob-grahn:
:+1:

PR #517, opened 2018-07-25
Prevent double player search

BrokenBulb78:
Adding `die();` will prevent outputting two player search input boxes and also prevent searching an empty name when you enter the page.

image


bls1999:
See #513

BrokenBulb78:
Ah, didn't see that. Thanks! @bls1999

PR #519, opened 2018-07-26
Convert warnings to mutes

bls1999:
This pull separates chat bans from the player object in favor of a server-logging system (like how kicking works). The difference will be negligible to a normal player, but this will allow us to do a lot more with chat bans than before, as well as abolishing warn evading by logging out and back into an account with an active warning.

In the future, with regards to mutes/warnings, I hope to:
 - Add a chat command that will allow a mod to mute a user for up to a day
 - Add an unwarn chat command (the function is already in `fns/client/moderation.php`)
 - Move all chat commands to a different php file so they aren't cluttering `Player.php`

This also fixes a bug with kicking where if a user attempts to log in using a different case as was previously used (i.e. JIGGMIN instead of Jiggmin), the kick doesn't have any effect.

### DIFF
--- a/multiplayer_server/Mutes.php
+++ b/multiplayer_server/Mutes.php
@@ -1,0 +1,83 @@
+<?php
+
+namespace pr2\multi;
+
+class Mutes
+{
+
+    private static $arr = array();
+
+
+    public static function add($user_name, $duration = 60)
+    {
+        $mute = new \stdClass();
+        $mute->user_name = $user_name;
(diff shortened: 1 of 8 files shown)


bls1999:
Travis Status for this branch (idk why it isn't showing on its own): [Build Status](https://travis-ci.org/jacob-grahn/platform-racing-2-server/builds/419968625)

jacob-grahn:
:+1:

PR #520, opened 2018-07-29
Make the links clickable in ban popup

BrokenBulb78:


bls1999:
Good pull! For future reference @BrokenBulb78:
 - Travis starts yelling when there's both logic and functions in the same file. The require link counts as logic, even though it's just calling another file.
 - Always use the complete URL with `urlify` (include the protocol, I.e. http/https/ftp)

BrokenBulb78:
Fixed! Thanks...

And about the missing protocol, I noticed that when I made the pull. (Because I copied and pasted the link, I MEAN wrote the link myself and forgot `https://`) :joy:

jacob-grahn:
:+1:

PR #523, opened 2018-09-02
Don't let campaign levels be deleted

bls1999:
Before deleting a level, the server will check to see if the level is in a Campaign. If it is, it won't let the user delete the level.

This has happened before. While it's not a common thing (it's only happened about 1-5 times in PR2's history), I think it's probably best to have this check to save you the trouble of having to restore it if it happens. Just in case; especially with new Campaign levels every month from Campaign of the Month!

(I also can't test this without making tons of dummy levels on my localhost, so just take a gander and make sure everything looks up to par.)

Also, please run this query on the database:

```mysql
UPDATE `pr2_levels`
  SET `live` = "1"
  WHERE `level_id` = 1442589;
```

EDIT: To be clear, this protects against a level being deleted via the `delete_level.php` script, meaning the creator cannot delete the level. The reason I didn't add this to the mods' `remove_level.php` is that `remove_level.php` only unpublishes the level, and the script that fills the Campaign ignores whether or not the level is published.

### DIFF
--- a/common/queries/campaign/campaign_level_select_by_id.php
+++ b/common/queries/campaign/campaign_level_select_by_id.php
@@ -1,0 +1,19 @@
+<?php
+
+function campaign_level_select_by_id($pdo, $level_id)
+{
+    $stmt = $pdo->prepare('
+        SELECT campaign, level_num
+        FROM pr2_campaign
+        WHERE level_id = :level_id
+        LIMIT 1
+    ');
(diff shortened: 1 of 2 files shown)


bls1999:
Travis Status for this branch (idk why it isn't showing on its own): [Build Status](https://travis-ci.org/jacob-grahn/platform-racing-2-server/builds/423549048)

PR #529, opened 2018-10-02
Improve ban output

BrokenBulb78:
Edits:
- Cleaned up code.
- Output **an IP** instead of blank text if the ban type is IP and you're a member.

### DIFF
--- a/http_server/www/bans/bans.php
+++ b/http_server/www/bans/bans.php
@@ -53,17 +53,13 @@ try {
         $formatted_time = date('M j, Y g:i A', $time);
         $duration = $expire_time - $time;
 
-        $display_name = '';
-        if ($account_ban == 1) {
-            $display_name .= $banned_name;
-        }
+        $display_name = $banned_name;
+
         if ($ip_ban == 1 && $is_mod) {
             if ($display_name != '') {
                 $display_name .= ' ';
             }
             $display_name .= "[$banned_ip]";
-        } elseif ($ip_ban == 1 && $account_ban == 0 && !$is_mod) {
-            $display_name = "<i>an IP</i>";
         }
 
         $reason = htmlspecialchars($reason);


EternalUF:
Waitttttttttttt, so if you are a member it will display someone else's IP instead of the usual blank?

bls1999:
Nah, not gonna give a thumbs up on this one. That information doesn't need to be on the ban log.

BrokenBulb78:
@EternalUF nah, it will display "an IP" lol

BrokenBulb78:
@bls1999 check Discord

BrokenBulb78:
At line 66 it re-defines the variable as `$display_name = "<b>an IP</b>";` so it wouldn't use htmlspecialchars on output, I might be wrong though.

bls1999:
Oh I brain farted and thought the red on 72/73 was the new code. That would be fine re: htmlspecialchars but there's still no need for `htmlspecialchars` on a blank string.

There are better ways to achieve this, but I still don't agree with the premise of the pull.

BrokenBulb78:
@bls1999 It's not actually blank since it adds the other text to it later, it looks like this `$display_name = htmlspecialchars("$banned_name [$banned_ip]");`

bls1999:
@Brokenbulb78 I'm talking about line 56. No need to escape with htmlspecialchars because there's nothing there.

bls1999:
@brokenbulb78 I know what you mean, but that's not how htmlspecialchars works. Putting htmlspecialchars around something only escapes that thing. You put it around a blank string, so it only applies to that string. The rest of the code just concatenates (adds to) the existing string. So basically, what you've made is this:

```php
$display_name = htmlspecialchars('') . $banned_name . ' ' . "[$banned_ip]";

**Comment by BrokenBulb78** [comment]
Oh I didn't know it worked like that LOL

I'm going to fix my mistakes later, thanks homie \o

**Comment by BrokenBulb78** [comment]
Wait I guess it's fine now since `$banned_name` has htmlspecialchars already

**Comment by bls1999** [comment]
I'm still not a fan of this pull just because I feel like it's not a good way to show the ban type. It would be much more effective to show it on show_record.php, if anywhere at all for members. Again, though, I don't think that information needs to be public.

**Comment by famelegend1** [comment]
dont think it's really that important in all honesty. who cares which bans are ip based and which from account?

**Comment by BrokenBulb78** [comment]
@famelegend1 it doesn't say which account has the IP, all it does is replacing the blank banned name with "an IP".

@bls1999 the information is public already and it's not hard to find out the ban type, the change just replaces the ugly blank text.

**Comment by BrokenBulb78** [comment]
@bls1999 owo
image

**Comment by bls1999** [comment]
The reason I was against it so much is that I thought it would say "[An IP]" on every ban.

Another option is to always display the name of the user on the ban log (regardless of ban type) and just list the type on show_record.php. I think that would be cleaner since, in this case, nothing (an "ugly" message) would still be displayed:

```php
$ip_ban = $row->ip_ban; // val = 0
$account_ban = $row->account_ban; // val = 0

$display_name = '';
if ($account_ban == 1) { // does not match conditions
    $display_name .= $banned_name;
    }
if ($ip_ban == 1 && $is_mod) { // does not match conditions
    if ($display_name != '') {
        $display_name .= ' ';
    }
    $display_name .= "[$banned_ip]";
} elseif ($ip_ban == 1 && $account_ban == 0 && !$is_mod) { // does not match conditions
    $display_name = "<i>an IP</i>";
}

echo "... $display_name ..."; // str = ""
```

Sure, we could make that if statement *even longer* with this pull, or we could just list the type in one place for members and remove the `elseif` block.

BrokenBulb78:
> Another option is to always display the name of the user on the ban log (regardless of ban type) and just list the type on show_record.php. I think that would be cleaner since, in this case, nothing (an "ugly" message) would still be displayed

I like this idea but can it output the banned person's name if the ban type is IP?

bls1999:
@BrokenBulb78 yep. You can't ban an IP without first specifying an account.

BrokenBulb78:
@famelegend1 a valid reason would be appreciated.

BrokenBulb78:
Uhh I deleted and re-forked the repo to clean up some mistakes and now it doesn't let me edit the commits. I'll close and re-create this pull.
image

PR #532, opened 2018-10-05
[Cosmetic] Improve ban output

BrokenBulb78:
Changes in **bans.php**:
- Always display the banned person's name regardless of ban type.
- Cleaned up a tiny piece of code.

Changes in **show_record.php**:
- The same ones in bans.php.
- Display the ban type after the reason.

### DIFF
--- a/http_server/www/bans/bans.php
+++ b/http_server/www/bans/bans.php
@@ -44,42 +44,38 @@ try {
         $banned_user_id = $row->banned_user_id;
         $time = $row->time;
         $expire_time = $row->expire_time;
-        $reason = $row->reason;
-        $mod_name = $row->mod_name;
-        $banned_name = $row->banned_name;
+        $reason = htmlspecialchars($row->reason);
+        $mod_name = htmlspecialchars($row->mod_name);
+        $banned_name = htmlspecialchars($row->banned_name);
         $ip_ban = $row->ip_ban;
         $account_ban = $row->account_ban;
 
         $formatted_time = date('M j, Y g:i A', $time);
         $duration = $expire_time - $time;
 
-        $display_name = '';
-        if ($account_ban == 1) {
-            $display_name .= $banned_name;
-        }
         if ($ip_ban == 1 && $is_mod) {
-            if ($display_name != '') {
-                $display_name .= ' ';
+            if ($banned_name != '') {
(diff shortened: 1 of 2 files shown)


famelegend1:
not needed. i dont think people care if a ban is ip or account. too much work for little reward.

famelegend1:
rather this stay untouched. you're overthinking this and putting too much in for hardly any change.

BrokenBulb78:
@famelegend1 the only reason I made this is because I don't like the blank name on IP ban. As @bls1999 said, it would be nicer to always display the banned name and include the ban type in show_record.php.
All it does is this:
image

bls1999:
@BrokenBulb78 
screen shot 2018-10-06 at 3 49 43 am

BrokenBulb78:
@bls1999 
image

bls1999:
@BrokenBulb78 refer to my previous comment

BrokenBulb78:
How so? It's much cleaner this time :I

bls1999:
I don't dispute that it's well-written, I just don't think it requires changing. That's just my opinion.

EternalUF:
Yeahhh, I agree this isn’t necessary. Typically when we do an IP only ban we state “IP ban” in the ban reason so in a way this is kinda redundant. Also, displaying the banned user’s name no matter the ban time would make the ban log cluttered in my opinion. This is a “not broke so don’t fix it” situation.

BrokenBulb78:
@EternalUF, the first idea was to just output "*an IP*" instead of the usual blank. I can change it to that again if you want to. I believe it would be less confusing. :)

bls1999:
@BrokenBulb78 you're missing the point. 3 people have said it doesn't need changing. It's not about the details or your code; we simply don't believe this is something the repo needs.

BrokenBulb78:
I don't understand why you guys prefer the blank text tho @bls1999

bls1999:
Because we believe it's fine the way it is. It's such a small thing.

BrokenBulb78:
@bls1999 3 is not a really big number, a poll would be nice. If the nos win, I'll probably transform this into a code cleaning pull. ¯\\\_(ツ)_/¯

BrokenBulb78:
¯\\\_(ツ)_/¯

PR #534, opened 2018-10-12
Display correct GP values in guild_search.php

Stxtics:
lol


jacob-grahn:
I had to re-read this change three times before I could actually see what the problem was

PR #538, opened 2018-11-09
Allow staff to purge a user's active login tokens

bls1999:
This pull lets permanent moderators and admins clear all stored login tokens for an account (to be used in case of an account security breach or other emergencies). For those that don't know, mods can already do this by banning. This script also includes rate limiting for performing this action on fellow staff members.

Additional [small] changes:
 - Quality of life updates for some of the admin tools (adding links to other scripts where there was formerly just plain text, i.e. IP addresses now link to `/mod/ip_info.php`)
 - Localization of the links found in `output_fns.php` (so that localhost testing is not the worst thing in the world if you accidentally click one of those links on the page).


EternalUF:
YESSS MUCH NEEDED

PR #539, opened 2018-11-13
Streamline private server purchases via VoM

bls1999:
In order to complete this pull request, you must run this command on the `servers` table:

```mysql
ALTER TABLE `servers`
  MODIFY `population` int(11) NOT NULL DEFAULT '0',
  MODIFY `status` varchar(10) NOT NULL DEFAULT 'down',
  MODIFY `active` tinyint(4) NOT NULL DEFAULT '1',
  MODIFY `salt` varchar(40) NOT NULL DEFAULT 'COMM_PASS', /* We could remove this; see below */
  MODIFY `guild_id` int(11) NOT NULL DEFAULT '0';
```

**Be sure to replace `COMM_PASS` with the actual server salt.** You can either do that or not even store that value in the db and just take all server salts from the `$COMM_PASS` variable in `env.php` since it never changes. If you want to go this route, go ahead and run this query and I can apply those changes in my next pull.

This pull also fixes server restarting so that `restart_server.php` and `restart_server()` actually restart the server as intended. What a concept!

Travis **staus**: [Build Status](https://travis-ci.org/jacob-grahn/platform-racing-2-server/builds/454793029)

### DIFF
--- a/common/queries/servers/server_insert.php
+++ b/common/queries/servers/server_insert.php
@@ -1,25 +1,19 @@
 <?php
 
-function server_insert($pdo, $server_name, $address, $port, $expire_timestamp, $salt, $guild_id)
+function server_insert($pdo, $expire_time, $server_name, $address, $port, $guild_id)
 {
-    $expire_date = date("Y-m-d H:i:s", $expire_timestamp);
     $stmt = $pdo->prepare('
         INSERT INTO servers
-           SET server_name = :server_name,
+           SET expire_date = FROM_UNIXTIME(:expire_time),
+               server_name = :server_name,
                address = :address,
                port = :port,
-               expire_date = :expire_date,
-               active = true,
-               salt = :salt,
-               guild_id = :guild_id,
-               population = 0,
-               staus = "down"
+               guild_id = :guild_id
     ');
+    $stmt->bindValue(':expire_time', $expire_time, PDO::PARAM_INT);
     $stmt->bindValue(':server_name', $server_name, PDO::PARAM_STR);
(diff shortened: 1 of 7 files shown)


EternalUF:
I mean sure if you wanna

common/queries/servers/server_insert.php
```diff
@@ -1,25 +1,19 @@
 <?php
 
-function server_insert($pdo, $server_name, $address, $port, $expire_timestamp, $salt, $guild_id)
+function server_insert($pdo, $expire_time, $server_name, $address, $port, $guild_id)
 {
-    $expire_date = date("Y-m-d H:i:s", $expire_timestamp);
     $stmt = $pdo->prepare('
         INSERT INTO servers
-           SET server_name = :server_name,
+           SET expire_date = FROM_UNIXTIME(:expire_time),
```
jacob-grahn: FROM_UNIXTIME doesn't seem to work here, for some reason. My small fix to get server rentals working was to add the `$expire_date = date("Y-m-d H:i:s", $expire_timestamp);` line that manually makes the timestamp -> date conversion before passing the value to the db

bls1999: I don't think `FROM_UNIXTIME` was the issue; I believe it was working fine when we converted the queries to PDO. Maybe try it first without merging into the repo? If it works, then it can be merged.

bls1999: @jacob-grahn ^^

bls1999:
Will include with Organipocalypse pt. 4.

PR #542, opened 2019-01-14
Sync server with new client + Organipocalypse pt. 4

bls1999:
This pull will bring the server up-to-date with the new version of the PR2 client I'm sending you. It also completes part 4 of the organipocalypse.

You'll need to modify your nginx settings so the server knows that the new `WWW_ROOT` is `http_server/` instead of `http_server/www/`.

Queries that need to be done in addition to this:

```mysql
-- QUERIES TO RUN: --

DELETE FROM pr2_campaign WHERE level_id = 3236908; # golden compass; this entry is made redundant by Game.php in multi
ALTER TABLE `recent_logins` ADD INDEX(`user_id`); # significantly decrease server load when calling data from this table
RENAME TABLE pr2_campaign TO campaigns;
RENAME TABLE pr2_levels TO levels;
RENAME TABLE pr2_new_levels TO new_levels;
RENAME TABLE pr2_ratings TO ratings;
DROP TABLE domains; # unused
DROP TABLE guilds_newer; # unused
DROP TABLE login_attempts; # truncate query exists, but otherwise unused
DROP TABLE users_new; # insert and delete old queries exist but are not used

ALTER TABLE `servers`
  MODIFY `population` int(11) NOT NULL DEFAULT '0',
  MODIFY `status` varchar(10) NOT NULL DEFAULT 'down',
  MODIFY `active` tinyint(4) NOT NULL DEFAULT '1',
  MODIFY `salt` varchar(40) NOT NULL DEFAULT 'COMM_PASS', # Make sure this reflects the actual COMM_PASS!
  MODIFY `guild_id` int(11) NOT NULL DEFAULT '0';
```

**Be sure to replace `COMM_PASS` with the actual server salt.** You can either do that or not even store that value in the db and just take all server salts from the `$COMM_PASS` variable in `env.php` since it never changes. If you want to go this route, go ahead and run the query anyway and I'll apply those changes in my next pull.


jacob-grahn:
Files changed: 491
*asplodes*
*approved*

PR #546, opened 2019-01-17
Fix allowed version

bls1999:
typos are fun...


BrokenBulb78:
Ben lives in the future confirmed

bls1999:
Actually... makes no sense to bother you with this when I can just change the version client-side before I send you the fixed version. :confused:

PR #558, opened 2019-02-21
you forgot to change shrug on /emotes

Stxtics:
I cannot believe that you changed it on one thing but not the other. I am very disappointed in you and because of this, I'm removing your bird set on PR2. I hope this is a lesson learned and it won't happen again. I expect a 500-word minimum letter from you explaining how sorry you are and some proof that this will not happen again, thanks. 

-Stxtics


bls1999:
‾\\\_(ツ)\_/‾

PR #560, opened 2019-02-22
Find values for future development

bls1999:


bls1999:
would help if I actually merged this! 🤦‍♂️

PR #587, opened 2019-04-25
Take off the 2 from player_info_2.php

BrokenBulb78:
Since the client has been decrypted and edits are possible I think the 2 can be taken off the file name now.
Merge once every file has been fixed in the client (or github repo, who knows ¯\\\_(ツ)\_/¯)


bls1999:
yeah, I can add this to my list of things to do for 156. thanks!

bls1999:
I went with a different method of changing this in order to aid backwards compatibility in the first few hours that 156 is released. thanks for the suggestion tho!

PR #590, opened 2019-05-03
Extend anniversary benefits through May 5

bls1999:
You see, the anniversary celebration doesn't really work if you don't restart the servers to apply the changes...


bls1999:
Alright... let's actually remember to restart the servers this time. 🤞

PR #592, opened 2019-05-12
Updating Shane's and Grovbi's profile

AshleyJ14:
Both of the links were broken. I went ahead and changed them to the correct one.


ProPlayzBlitzCrank:
Pro coder

PR #618, opened 2019-11-15
Update email

jacob-grahn:
I think this may be the missing sauce in getting email deliverability sorted out. Mailgun is set up to use mg.pr2hub.com, because their docs suggested suggested not using the main domain for some reason


bls1999:
Interesting! This is very possible. I believe the error stems from the "from" address's displayed domain not matching the sending IP. A lack of MX records on PR2 Hub can also cause this.

bls1999:
It works! Great sleuthing @jacob-grahn! Thanks for everything. :)

PR #685, opened 2020-11-25
Tidy up strToId in Parts.php

Northadox:
So I was browsing through this repository once upon a time and I saw that in the function "strToId" in Parts.php that with the Jump Start Hat, instead of having a bunch of "or" conditionals, it just searched through an array. Array $jump_start was declared beforehand to do this, and no such thing occurred with no other hat. So just today while looking through it again I thought "why not do it for all hats, might make it tidier?" So I did. I also added the string "arti" in the checks for the Artifact, 'cuz that's a very common shorthand for the hat.

Now I'm not sure if this is what you'd want. Maybe this is some ancient Jiggy code you didn't want to interfere with for prosperity's sake. Maybe the Jump Start hat must have an array declared when converting a string to its ID or else Fred grows irritable. Honestly, I don't know a lick about PHP - I just wrote this through extrapolation of the logic and syntax. Yes, I checked through the linked site, it works, no syntax errors. From a programmer's point of view, maybe some sort of switch statement would be best for such a function, but I wouldn't know what works for you. Could make it easier for the future, for adding more hats and whatnot. It does look a bit more legible to me but again it's called a "pull request" not a "pull demand". New to this version control stuff, it's really neat.

Regards,
Northadox

### DIFF
--- a/multiplayer_server/parts/Parts.php
+++ b/multiplayer_server/parts/Parts.php
@@ -69,39 +69,55 @@ class Hats
     {
         $str = strtolower($str);
         $id = 1;
-        $jump_start = ['start', 'jump', 'jumpstart', 'jump_start', 'js', 10];
+		
+		$none = ['none', 'n', '', 1];
+		$exp = ['exp', 'experience', 'e', 2];
+		$kong = ['kong', 'kongregate', 'k', 3];
+		$propeller = ['propeller', 'prop', 'pr', 4];
+		$cowboy = ['cowboy', 'gallon', 'co', 5];
+		$crown = ['crown', 'cr', 6];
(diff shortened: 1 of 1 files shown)


bls1999:
Nice idea! I'm always in the boat for simplifying code. There's also the added bonus of someone other than me contributing to this repo. :)

I believe the original purpose of the simple `if` conditions was for speed. I know `in_array` is slower than that, but I'm not sure how much slower or if it matters at all. I'll run a few tests and let you know.

jacob-grahn:
Looks good to me 👍 

If you want to be even fancier, you could make an array of your arrays and loop through. When you find a match, set $id to the last element of the inner array and break out of the loop.

PR #687, opened 2020-11-26
Update to StrToId in Parts.php

CamerTheDragon:
- Moved the jump_start element to the start of the jumpstart hat's array and the top_hat element to the start of the top hat's array (this was because I was originally going to set the id with the 1st element of the array as upper case but then I realised I could just do it with the last element)
- Put all the arrays in 1 big array
- Made a foreach loop that goes through the array that checks whether the string is in each of the sub-arrays
- If it's in the sub-array, then the id is the last element of the sub-array

It didn't say anything was wrong in code checker for both versions 7.3 and 5.6 but I've never used php before so there's a good chance that something here is wrong

EDIT: IdToStr too

### DIFF
--- a/multiplayer_server/parts/Parts.php
+++ b/multiplayer_server/parts/Parts.php
@@ -69,40 +69,13 @@ class Hats
     {
         $str = strtolower($str);
         $id = 1;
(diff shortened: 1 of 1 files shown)


bls1999:
I'll review this tonight

PR #689, opened 2020-12-06
Small edit to idToStr and getDesc

CamerTheDragon:
Changed the manual 16 in the if statements to be the length of the array instead (so it updates with the number of hats by itself)


bls1999:
Looks good to me! 👍🏻

Issue #731, opened 2022-01-03
error running liquibase for the first time


fourst4r:
this prevents the database from being set up properly
```
liquibase_1  | Unexpected error running Liquibase: Validation Failed:
liquibase_1  |      2 change sets had duplicate identifiers
liquibase_1  |           liquibase/base.mysql.sql::1574908323264-70::root
liquibase_1  |           liquibase/base.mysql.sql::1574908323264-73::root
liquibase_1  |
liquibase_1  | For more information, please use the --logLevel flag
liquibase_1 exited with code 255
```

PR #733, opened 2022-02-28
[Potentially] Fix Liquibase

bls1999:
This contains a potential fix for #731.


bls1999:
@fourst4r please lmk if this works

fourst4r:
this is still a problem @bls1999 
```
docker-liquibase-1  | Unexpected error running Liquibase: Validation Failed:
docker-liquibase-1  |      1 change sets had duplicate identifiers
docker-liquibase-1  |           liquibase/base.mysql.sql::1574908323264-70::root```

fourst4r:
i fixed it by picking a new ID manually for one of the 70s but idk if that's the ideal solution, might be better to generate an updated schema from the real DB?

VipaPR2:
> 

I agree with this.

PR #730, opened 2022-01-03
Create .gitattributes

fourst4r:
make sure git doesn't convert the line endings or the shell scripts wont run properly on docker for windows

PR #734, opened 2022-03-01
Make docker-friendly (pt. 2)

bls1999:
This is related to #730. Since I don't use Docker, I'm looking for input from @fourst4r.

7efce6a is being weird. It seems like a better way to generate `http_server/files/server_status_2.txt` is to just run the minute CRON job at runtime. This combines a bunch of processes, including starting the existing servers.


bls1999:
I'm gonna go ahead and merge this.

fourst4r:
so this doesn't work because the database is set up last, so the cron runs and can't access the DB and dies
```
docker-web-1  | * Minute CRON starting at Mon, 18 Jul 2022 12:10:32 +0000...
docker-web-1  | 
docker-web-1  | PHP Fatal error:  Uncaught Exception: Could not connect to the database. in /pr2/common/pdo_connect.php:20
docker-web-1  | Stack trace:
docker-web-1  | #0 /pr2/common/cron/minute.php(17): pdo_connect()
docker-web-1  | #1 {main}
docker-web-1  |   thrown in /pr2/common/pdo_connect.php on line 20
```

fourst4r:
Not sure if you get notified of comments on this so I'll just @bls1999

PR #756, opened 2023-09-23
Fix botting check (again).

darkeverlast:
Prevents a false positive of playing lag test levels with jump start.


bls1999:
What's the level at which the issue was observed?

darkeverlast:
> On what level was this issue observed?

Create a level with a finish block and negative gravity and enter the level with jump start hat (with the player start points being just below the finish blocks). This is used to check your finish lag. If you have lower ping to the server, this finish lag will be less (let's say, 0.3s). If you have higher ping, it will be more.

The problem is when this type of level is played multiple times very quickly, the rate limiting treats this as a false positive and disconnects you from the server. This fixes that, and potentially fixes another issue (item farming with jump start hat).

